Issue #11, opened 2018-05-03
Access Control


brianteeman:
In theory (and especially if you are using role based acl) you could have a site with the following scenario

- admin with only access to the ecommerce component
- admin with only access to the newsletter component
- admin with only access to the membership component

For the privacy component to work you would need to have a user with access to all the information otherwise someone with only newsletter access would see data that they were not intended to see from the ecommerce component

So "maybe" the solution is to restrict access to this component to super users like akeeba does

brianteeman:
Another real world example would be on joomla.org. The volunteer portal says 

> According to section 4.03 of the Open Source Member By-laws your address is required as Member of a team. The provided mailing address is only **visible for the Secretary of Open Source Matters** and only required for active members of a team.

peteruoi:
Maybe a new predefined role DPO between administrator and super administrator?

brianteeman:
I wouldnt do it that way but use a role that can be applied to a specific user or users. You many not want to give the DPO access to other aspects of the site so a hierarchical option would not be appropriate

mbabker:
> So "maybe" the solution is to restrict access to this component to super users like akeeba does

My initial thought was to go for that too.  Because exactly as you pointed out, there are a massive number of scenarios to take into account where someone without read access to another portion of the site could see some sensitive data, therefore the safest option would be restricting to super users who should already be entrusted to see all the things.

JoomliC:
I will go too for superuser (Controller ?) as default VS others with admin access (Processor?).

Maybe a related information about this point in GDPR, Article 18 - Right to restriction of processing:  
https://gdpr-info.eu/art-18-gdpr/

We can then maybe just have additional access option like this :  
`<action name="core.privacy" title="JACTION_PRIVACY" description="JACTION_PRIVACY_COMPONENT_DESC" />`
Which one could be set in each 3rd party component?

zero-24:
Implementation done here: https://github.com/joomla-projects/privacy-framework/pull/13

PR #13, opened 2018-05-03
Implement access checks as discussed in #11

zero-24:
Pull Request for Issue #11

### Summary of Changes

The Backend component can only be opend by a super user and the frontend only authenticated users 

### Expected result

The Backend component can only be opend by a super user and the frontend only authenticated users 

### Actual result

The Backend component can only be opend by a any user with core.manage and the frontend can be opend by everyone

none


mbabker:
Might as well clean out `access.xml` and the config reference as well 😉

zero-24:
I have left the config file empty as i guess we would need that file later for the options ;)

Issue #9, opened 2018-05-03
Tracking Information Requests


mbabker:
Something that is going to help site administrators is an ability to track information requests (and this will tie in pretty closely to #3 and possibly other actions as things go).  We should have an interface for tracking the status of requests (generically this can cover requests for a user's information and requests to be deleted as a starting point) and as tools build up around it integrate those into the interface (i.e. if we are able to provide an automated download for #3 then the action for this would come from this view).

Note that this should be able to support requests for users with accounts on the site and what would essentially be guest accounts (as there would be ways for an unauthenticated user to interact with the site and have data in place falling under a rightful request to be forgotten/deleted, such as comments on a blog post).

WIP feature branch for this action: `feature/requests-list`

brianteeman:
it will be important that these requests are permanently logged

for example
person x requests their information be deleted on 1 march
we reply that it cant be because of legal reasons such as tax reporting
x requests again on 1 april, 1 may, 1 june etc

mbabker:
I don't intend on exposing a delete button in this context.

Once completed, information such as a request's status, created date, acted on person/date (essentially whomever marked a request as invalid, complete, or something that would equate to a "closed" workflow transition) will be included in this list.

mbabker:
Migrated to #15

PR #15, opened 2018-05-04
[Feature] Tracking Information Requests

mbabker:
Pull Request for Issue #9

### Summary of Changes

This will be the central piece of the user workflow for managing information requests.  Not entirely "complete", but functional enough to introduce a PR and share the methodology and approach here.

### Testing Instructions

1) Download this branch
2) Install Joomla
3) Navigate to `Components > Privacy`
4) Interact with component

Absolutely 😉

#### General Notes

- Purposefully this system will NOT support deleting requests (if this really becomes burdensome we can revisit later but this screen can act as an audit log and we shouldn't allow that data to be dropped easily)

#### Screenshots

Main list view
screen shot 2018-05-06 at 1 13 54 pm

Add request screen
screen shot 2018-05-06 at 1 12 02 pm

Show request screen
screen shot 2018-05-06 at 1 14 16 pm

#### Requests Workflow

This PR implements one aspect of the data request workflow, the ability for the site administrator to log a request.  As it is likely a request will come from someone who does not have an account on the site, the identifier for these requests will be the user's email address; optionally this request MAY be linked to a user account.

Requests use a workflow style model for statuses which looks something like this:

- New Item
    - To: Pending
- Request Validated
    - From: Pending
    - To: Confirmed
- Request Completed
    - From: Confirmed
    - To: Completed
- Request Invalidated
    - From: Pending, Confirmed
    - To: Invalid

All requests start in a pending status.  GDPR requires identity verification for these types of requests, so requests in a pending status indicates that this verification has not yet been completed (for now this will be in the form of an email verification similar to password resets, but this isn't implemented yet).  A request can transition to either Confirmed or Invalid from here; Confirmed would indicate the user responding to the verification email and Invalid would indicate that the request timed out (TODO - Have a system plugin or CLI script that can be cron'ed that will check for pending requests older than X hours, presumably something like 24, 48, or 72, and invalidate them).  In a Confirmed status, either it can be acted upon (Completed) or the site admin can invalidate the request.  As indicated in the above list, a request in an Invalid or Completed status cannot be transitioned.

### TODO (not necessarily blockers to the merge of this PR but definitely work to be continued)

- [x] Disallow creating a request for the currently authenticated user
- [x] Disallow creating a new request for a user who has an active request (status of Pending or Confirmed), we should nudge admins to work with these active requests
- [ ] Think about a Failed status in addition to Invalid
    - This would be the "user never confirmed validation" status and could potentially give admins the possibility to retry validation if it expires (up to a sane limit)
- [ ] Think about some kind of help text directly on the single request layouts
    - This is a bit of an odd case, requests are basically immutable once received outside the status field so there may need to be more steering of users as to why this section works this way and what to do if they do make a mistake, something for some UX folks to consider
- [x] Support multiple request types (this can just be a new `request_type` column for now supporting "export" and "removal" types, leave this flexible enough to add more types if need be)
- [x] Implement post save hook for new items to send out the verification email


sandewt:
Great, I have been able to reproduce the sreenshots. 

Missing link in componts to `administrator/index.php?option=com_privacy` yet.

mbabker:
Unless I screwed something up in https://github.com/joomla-projects/privacy-framework/commit/4b0f1c1bbb314bf6d9377f8187f9549d4158de9f all the menu scaffolding should've been added (either way that's outside scope of this PR, this commit was made to the repo to have the component skeleton in place to make PRs against).

administrator/components/com_privacy/models/forms/filter_requests.xml
```diff
@@ -0,0 +1,51 @@
...
+
+	<fields name="list">
+		<field
+			name="fullordering"
+			type="list"
+			label="JGLOBAL_SORT_BY"
+			description="JGLOBAL_SORT_BY"
+			onchange="this.form.submit();"
+			default="a.id DESC"
+		>
```
brianteeman: indent one tab

mbabker:
OK, aside from needing the post-create email hook and probably some UX love fundamentally this is ready for testing/review.

brianteeman:
Seems to do what you intended it to do without error

brianteeman:
I can't find a reference for this but my gut feeling is that you say
Requested at 2:30
Requested on 1st July
Requested on 1st July at 2:30

Requested at 2018-01-01 09:30 just seems wrong

but its a minor thing

mbabker:
Grammatically speaking we'd need a `DATE_FORMAT_LC7="j F Y \a\t H:i:s"` string to do it right (turns into "1 December 2017 at 19:30:00").  Lacking that, what's there now seems to be the best option.

brianteeman:
Or avoid the problem completely and change the field from "Requested at" to "Date Requested"

mbabker:
Label changed.

mbabker:
Added in the hook for sending the confirmation email out.  I'm done for this PR, at this point everything to start work on frontend elements is in place, there's now a TODO in code that needs us merging the branch with https://github.com/joomla/joomla-cms/pull/16879 merged over here.

mbabker:
I'm content with where this is at.  Merging to allow next phase to be built out.

PR #20, opened 2018-05-05
[Action logs] - component+plugin+module

alikon:
Pull Request for Issue #4 .

### Summary of Changes
first step of porting https://github.com/joomla/joomla-cms/pull/11834

TO DO

- [x] The language strings can be massively simplified / rivisited
- [x] mssql files
- [x] code cleaning
- [x] 3dp extensions support see #38 for an how-to
- [x] sql to install with sample data


administrator/components/com_userlogs/helpers/userlogs.php
```diff
@@ -0,0 +1,136 @@
...
+	 * @param   array  $data  Has the data to be exported
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public static function dataToCsv($data)
+	{
+		$date = JFactory::getDate();
+		$filename = "Logs_" . $date;
```
brianteeman: because we can get into issues with case sensitivity and this file doesn't need it can you make the name all lowercase

administrator/components/com_userlogs/helpers/userlogs.php
```diff
@@ -0,0 +1,136 @@
...
+		rewind($fp);
+		$content = stream_get_contents($fp);
+		echo $content;
+		fclose($fp);
+
+		$app->close();
+	}
+
+	/**
+	 * Change the retrived extension name to more user friendly name
```
brianteeman: retrieved

administrator/components/com_userlogs/helpers/userlogs.php
```diff
@@ -0,0 +1,136 @@
...
+		if (empty($items))
+		{
+			return false;
+		}
+
+		return $items[0];
+	}
+
+	/**
+	 * Method to retrive data by primary keys from a table
```
brianteeman: retrieve

administrator/components/com_userlogs/models/userlogs.php
```diff
@@ -0,0 +1,335 @@
...
+
+			case 'past_3month':
+				$dStart->modify('-3 month');
+				break;
+
+			case 'past_6month':
+				$dStart->modify('-6 month');
+				break;
+
+			case 'post_year':
```
brianteeman: is this one correct? post?

administrator/components/com_userlogs/helpers/userlogs.php
alikon: lowered

administrator/components/com_userlogs/models/userlogs.php
alikon: should be better now

administrator/components/com_userlogs/userlogs.xml
```diff
@@ -1,36 +1,14 @@
...
 	<author>Joomla! Project</author>
-	<creationDate>June 2017</creationDate>
-	<copyright>Copyright (C) 2005 - 2017 Open Source Matters. All rights reserved.</copyright>
+	<creationDate>May 2018</creationDate>
+	<copyright>Copyright (C) 2005 - 2018 Open Source Matters. All rights reserved.</copyright>
 	<license>GNU General Public License version 2 or later; see LICENSE.txt</license>
 	<authorEmail>admin@joomla.org</authorEmail>
 	<authorUrl>www.joomla.org</authorUrl>
-	<version>0.0.1</version>
+	<version>3.0.0</version>
```
zero-24: should be __DEPLOY_VERSION__ or 3.9.0 :P

administrator/components/com_userlogs/sql/install.mysql.sql
```diff
@@ -0,0 +1,64 @@
+CREATE TABLE IF NOT EXISTS `#__user_logs` (
```
zero-24: please move that files to the core install sql ;)

administrator/components/com_userlogs/sql/unistall.mysql.sql
```diff
@@ -0,0 +1,3 @@
+DROP TABLE IF EXISTS `#__user_logs`;
+DROP TABLE IF EXISTS `#__user_logs_extensions`;
```
zero-24: you can drop this file

administrator/components/com_userlogs/sql/unistall.postgresql.sql
```diff
@@ -0,0 +1,92 @@
+-- install.mysql.sql
+--
+-- Table structure for table `#__user_logs`
+--
+
+CREATE TABLE IF NOT EXISTS `#__user_logs` (
```
zero-24: this seams to be a c& p issue but can be droped anyway.

administrator/components/com_userlogs/sql/unistall.sqlsrv.sql
```diff
@@ -0,0 +1,92 @@
+-- install.mysql.sql
+--
+-- Table structure for table `#__user_logs`
+--
+
+CREATE TABLE IF NOT EXISTS `#__user_logs` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
```
zero-24: same here drop it

administrator/components/com_userlogs/sql/updates/mysql/0.0.2.sql
```diff
@@ -0,0 +1,92 @@
+-- install.mysql.sql
+--
+-- Table structure for table `#__user_logs`
+--
+
+CREATE TABLE IF NOT EXISTS `#__user_logs` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `message` text NOT NULL DEFAULT '',
+  `log_date` datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
```
zero-24: as this is the initial version we should merge that files with the core install sql

administrator/components/com_userlogs/userlogs.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extension type="component" version="3.9" method="upgrade">
+	<name>com_userlogs</name>
+	<author>Joomla! Project</author>
+	<creationDate>May 2018</creationDate>
+	<copyright>Copyright (C) 2005 - 2018 Open Source Matters. All rights reserved.</copyright>
+	<license>GNU General Public License version 2 or later; see LICENSE.txt</license>
+	<authorEmail>admin@joomla.org</authorEmail>
+	<authorUrl>www.joomla.org</authorUrl>
+	<version>3.0.0</version>
```
zero-24: 3.9.0 or `__DEPLOY_VERSION__` ;)

administrator/components/com_userlogs/userlogs.xml
```diff
@@ -0,0 +1,31 @@
...
+		<files folder="admin">
+			<file>userlogs.php</file>
+			<file>config.xml</file>
+			<file>access.xml</file>
+			<file>controller.php</file>
+			<folder>controllers</folder>
+			<folder>helpers</folder>
+			<folder>models</folder>
+			<folder>views</folder>
+			<folder>sql</folder>
```
zero-24: sql should be dropped.

administrator/components/com_userlogs/views/userlogs/view.html.php
```diff
@@ -0,0 +1,107 @@
...
+	 * @var  array
+	 *
+	 * @since __DEPLOY_VERSION__
+	 */
+	protected $items;
+
+	/**
+	 * The model state
+	 *
+	 * @var  object
```
zero-24: since tag missing

administrator/components/com_userlogs/views/userlogs/view.html.php
```diff
@@ -0,0 +1,107 @@
...
+	 */
+	protected $state;
+
+	/**
+	 * The pagination object
+	 *
+	 * @var  JPagination
```
zero-24: since tag

administrator/components/com_userlogs/views/userlogs/view.html.php
```diff
@@ -0,0 +1,107 @@
...
+	 */
+	protected $pagination;
+
+	/**
+	 * The active search filters
+	 *
+	 * @var  array
```
zero-24: since tag

plugins/system/userlogs/layouts/logstable.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * @package	 Joomla.Administrator
+ * @subpackage  com_userlogs
+ *
+ * @copyright   Copyright (C) 2005 - 2016 Open Source Matters, Inc. All rights reserved.
+ * @license	 GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('_JEXEC') or die;
+$lang = JFactory::getLanguage();
```
zero-24: clean line please

plugins/system/userlogs/userlogs.php
```diff
@@ -0,0 +1,851 @@
...
+	/**
+	 * On deleting user data logging method
+	 *
+	 * Method is called after user data is deleted from the database
+	 *
+	 * @param   array    $user     Holds the user data
+	 * @param   boolean  $success  True if user was succesfully stored in the database
+	 * @param   string   $msg      Message
+	 *
+	 * @return  boolean
```
zero-24: since tag please

plugins/system/userlogs/userlogs.php
```diff
@@ -0,0 +1,851 @@
...
+	/**
+	 * On after save user group data logging method
+	 *
+	 * Method is called after user data is deleted from the database
+	 *
+	 * @param   string   $context  The context
+	 * @param   JTable   $table    DataBase Table object
+	 * @param   boolean  $isNew    Is new or not
+	 *
+	 * @return  boolean
```
zero-24: since tag

plugins/system/userlogs/userlogs.php
```diff
@@ -0,0 +1,851 @@
...
+	 * On deleting user group data logging method
+	 *
+	 * Method is called after user data is deleted from the database
+	 *
+	 * @param   array    $group    Holds the group data
+	 * @param   boolean  $success  True if user was succesfully stored in the database
+	 * @param   string   $msg      Message
+	 *
+	 * @return  boolean
+	 */
```
zero-24: since tag

plugins/system/userlogs/userlogs.php
```diff
@@ -0,0 +1,851 @@
...
+	}
+
+	/**
+	 * Method is called before writing logs message to make it more readable
+	 *
+	 * @param   string  $message    Message
+	 * @param   string  $extension  Extension that caused this log
+	 *
+	 * @return  boolean
+  */
```
zero-24: sincetag & intention ;)

administrator/components/com_userlogs/userlogs.xml
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<extension type="component" version="3.9" method="upgrade">
```
zero-24: please switch version & type ;)

administrator/components/com_userlogs/controller.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_userlogs
+ *
+ * @copyright   Copyright (C) 2005 - 2017 Open Source Matters, Inc. All rights reserved.
```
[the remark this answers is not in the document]
alikon: done

administrator/components/com_userlogs/userlogs.xml
zero-24: `<extension type="component" version="3.9" method="upgrade">`

vs

`<extension version="3.9" type="component" method="upgrade">`

brianteeman:
Is it ready for testing? Other than installing it i dont see any logging

zero-24:
@brianteeman did you enabled the plugin too?

brianteeman:
my error

brianteeman:
The plugin was enabled but I guess the default values were not correct as I had to save the plugin to write the params and then it worked

brianteeman:
The language strings can be massively simplified

For example
+PLG_SYSTEM_USERLOGS_ON_CONTENT_CHANGE_STATE_UNPUBLISHED_MESSAGE="%s(s) [\"%s\"] was unpublished"

This produces
`Article(s) ["test"] was unpublished
`

1. it doesnt need the Article(s) as we know it is the articles component because of the column COM_USERLOGS_EXTENSION	

2. when multiple articles were unpublished at the same time you get 
`	Article(s) ["meny\",\"test"] was unpublished
`
(ignoring the bug about the extra \)
this is bad english and probably most other languages because as a plural it should be "were unpublished"

So let's simplify things by just doing
[title] - [change]
this would producce
[articl1] - Unpublished
[articl1, articl2] - Published

I believe it will also make it much easier for translators

administrator/components/com_userlogs/models/forms/filter_userlogs.xml
```diff
@@ -0,0 +1,71 @@
...
+			<option value="">JGLOBAL_SORT_BY</option>
+			<option value="a.id ASC">JGRID_HEADING_ID_ASC</option>
+			<option value="a.id DESC">JGRID_HEADING_ID_DESC</option>
+			<option value="a.message ASC">COM_USERLOGS_MESSAGE_ASC</option>
+			<option value="a.message DESC">COM_USERLOGS_MESSAGE_DESC</option>
+			<option value="a.log_date ASC">JDATE_ASC</option>
+			<option value="a.log_date DESC">JDATE_DESC</option>
+			<option value="a.extension ASC">COM_USERLOGS_EXTENSION_ASC</option>
+			<option value="a.extension DESC">COM_USERLOGS_EXTENSION_DESC</option>
+			<option value="a.user_id ASC">JAUTHOR_ASC</option>
```
brianteeman: This isnt the correct string as in the context of publishing a module you would say user not author
so you will need a new string
COM_USERLOGS_NAME_ASC
or
COM_USERLOGS_USERNAME_ASC

depending if this is the real name or the username

brianteeman:
In the table view we can sort by ID but there is no column displayed for ID

We usually have ID as the last column and for the sort filters they are ordered to match the column orders so they will need to be reordered

brianteeman:
There is an option if you should log the ip address or not. In other places where we have a similar option then we don't display the column or filter

mbabker:
Looks like you're missing SQL to add the component to the extensions table too (use extension ID 36 since I used 35 for `com_privacy`).

administrator/components/com_userlogs/models/userlogs.php
```diff
@@ -0,0 +1,325 @@
...
+					->from($db->quoteName('#__user_logs', 'a'));
+			$db->setQuery($query);
+
+			return $db->loadObjectList();
+		}
+		else
+		{
+			$items = array();
+			JTable::addIncludePath(JPATH_ADMINISTRATOR . '/components/com_userlogs/tables');
+			$table = $this->getTable('Userlogs', 'JTable');
```
mbabker: Since you're not returning the `JTable` objects but a subset of records seems to me you're better off running one SELECT query with a WHERE IN clause versus this which is an N + 1 scenario.

administrator/components/com_userlogs/models/userlogs.php
```diff
@@ -0,0 +1,325 @@
...
+		// Check for request forgeries
+		JSession::checkToken() or die(JText::_('JINVALID_TOKEN'));
+
+		// Get the table
+		JTable::addIncludePath(JPATH_ADMINISTRATOR . '/components/com_userlogs/tables');
+		$table = $this->getTable('Userlogs', 'JTable');
+
+		if (!JFactory::getUser()->authorise('core.delete', $this->option))
+		{
+			$error = $this->getError();
```
mbabker: Entire if block can be simplified to just the `$this->setError()` and return lines

administrator/components/com_userlogs/userlogs.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * @package     Joomla.Administrator
+ * @subpackage  com_userlogs
+ *
+ * @copyright   Copyright (C) 2005 - 2016 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('_JEXEC') or die;
+JHtml::_('behavior.tabstate');
```
mbabker: Please don't arbitrarily load JS behaviors here.

administrator/components/com_userlogs/views/userlogs/tmpl/default.php
```diff
@@ -0,0 +1,115 @@
...
+							<td>
+								<span class="sortable-handler inactive tip-top hasTooltip">
+									<i class="icon-menu"></i>
+								</span>
+							</td>
+							<td class="center">
+								<?php echo JHtml::_('grid.id', $i, $item->id); ?>
+							</td>
+							<td>
+								<?php JEventDispatcher::getInstance()->trigger('onLogMessagePrepare', array (&$item->message, $item->extension)); ?>
```
mbabker: Is there a reason for dispatching the event in the layouts and within a loop versus in the view class?

plugins/system/userlogs/userlogs.php
```diff
@@ -0,0 +1,851 @@
...
+	 *
+	 * @param   string   $extension  The extension that triggered the event
+	 *
+	 * @return  boolean
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function checkLoggable($extension)
+	{
+		if (!in_array($extension, $this->loggableExtensions))
```
mbabker: Simplify, `return in_array();`

plugins/system/userlogs/userlogs.php
```diff
@@ -0,0 +1,851 @@
...
+		if (!$this->checkLoggable($context))
+		{
+			return;
+		}
+
+		$extensionName = (array) $installer->get('manifest')->name;
+		$message = array(
+			'event'          => 'onExtensionAfterUninstall',
+			'extension_name' => $extensionName[0],
+			'extension_type' => $installer->get('manifest')->attributes()['type'],
```
mbabker: This line's not PHP 5.3 compatible IIRC.

administrator/components/com_userlogs/views/userlogs/view.html.php
alikon: done

administrator/components/com_userlogs/views/userlogs/view.html.php
alikon: done

administrator/components/com_userlogs/views/userlogs/view.html.php
alikon: done

administrator/components/com_userlogs/models/userlogs.php
alikon: cleaned a bit

administrator/components/com_userlogs/models/userlogs.php
alikon: cleaned a bit

administrator/components/com_userlogs/userlogs.xml
alikon: done

administrator/components/com_userlogs/userlogs.php
alikon: removed

plugins/system/userlogs/userlogs.php
alikon: not sure what can be a PHP 5.3  compatible substitute
please help

plugins/system/userlogs/userlogs.php
alikon: done

administrator/components/com_userlogs/views/userlogs/tmpl/default.php
alikon: i'll look at this....

alikon:
it should be ready for a round of test 
sorry in advance for all the the glitch

alikon:
> When the option to log the ip address is disabled then the key PLG_SYSTEM_USERLOGS_DISABLED is written to the ip_address field log table - shouldnt it just be empty?

...lang string error apart..
in this way remind you that at that time your settings was to not collect ip data

alikon:
need help on language matter...
please help me change  langs on `feature/logging` branch should be open ....

brianteeman:
I would also add a purge button to the toolbar similar to com_redirect

brianteeman:
I changed some text strings for state  …
Works perfectly for single items
	[test] - archived

Not sure how to fix the multiple items
its somethign to do with the json_encode

[meny\",\"test] - unpublished

alikon:
the multiple actions logs as is it seems a little bit buggy to me
it should report as many log as many items actions....

brianteeman:
do you mean it should be one record in the log for each item? If so then I thought I agreed with that but then I realised that would generate a hundred emails if you did a batch publish of a 100 articles

plugins/system/userlogs/userlogs.php
mbabker: IIRC you can't direct access an array's attributes from a method call until PHP 5.5.

So you'd need an `$attributes = $installer->get('manifest')->attributes();` line then change this to `$attributes['type']`

joomdonation:
@mbabker As we removed JEventDispatcher in Joomla 4, should we use $app->triggerEvent instead of JEventDispatcher trigger?

mbabker:
That'd be better, yes (especially as `$app->triggerEvent()` has existed for a while and in 3.x is for the most part a fancy wrapper around `JEventDispatcher::getInstance()->dispatch()`.

joomdonation:
OK, thanks. Include that changes in my PR. Still few more questions:

1. Do we really need to have View Logs as a permission setting? As I can see, we only have that one view in the component, so maybe Access Administration Interface already cover that permission?

2. About permission checking. Right now, we are checking delete permission in model, shouldn't it being checked inside controller delete method? Same about checking View Logs permission, the code for checking it is inside View class now. Should it be checked inside display method of controller?

plugins/system/userlogs/userlogs.php
joomdonation: This is addressed in my PR.

```php
$manifest = $installer->get('manifest');
$message  = array(
			'event'          => 'onExtensionAfterInstall',
			'extension_name' => (string) $manifest->name,
			'extension_type' => (string) $manifest->attributes()->type,
		);
```

administrator/components/com_userlogs/views/userlogs/tmpl/default.php
joomdonation: I addressed it in my PR. Moved the code to view class.

alikon:
@joomdonation 
point 1. is debatable imo
point 2. can you please do a pr for cleaning/standardizing

@brianteeman 

> do you mean it should be one record in the log for each item? If so then I thought I agreed with that but then I realised that would generate a hundred emails if you did a batch publish of a 100 articles

one record in the log for each item, yes,  but only one email with the detailed report of items when `onContentChangeState`  ,btw, the `logs_notification_option` need to be reviewed

brianteeman:
one log and one email sounds fine to me

plugins/system/userlogs/userlogs.php
```diff
@@ -0,0 +1,846 @@
...
+			$this->db->quote($context),
+			$this->db->quote($user->id),
+			$this->db->quote($ip),
+		);
+
+		$query->insert($this->db->quoteName('#__user_logs'))
+			->columns($this->db->quoteName($columns))
+			->values(implode(',', $values));
+
+		$this->db->setQuery($query);
```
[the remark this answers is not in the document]
Septdir: Like that
```
// Prepare insert object
$log = new stdClass();
$log->message     = $json_message;
$log->log_date   = $date;
$log->extension  = $context;
$log->user_id    = $user->id;
$log->ip_address = $ip;

$this->db->insertObject('#__user_logs', $log);
```

joomdonation:
@mbabker In current code base, before log items being displayed, we call checkin method (wrong name, BTW) to delete records older than certain days https://github.com/joomla-projects/privacy-framework/blob/feature/logging/administrator/components/com_userlogs/models/userlogs.php#L77
That doesn't look right to me. Do you know where that function should be implemented?

alikon:
yes the name `checkIn` is really ugly...
do you mean in the plugin ? and/or in a CLI script ?

joomdonation:
Yes. Personal I think it should be in the plugin which will only be processed every X hours/days.... (like the update notification plugin). It should not be in the process of displaying logs as how we are doing at the moment

administrator/components/com_admin/sql/updates/mysql/3.9.0-2018-05-05.sql
```diff
@@ -0,0 +1,80 @@
...
+CREATE TABLE IF NOT EXISTS `#__user_logs_tables_data` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `type_title` varchar(255) NOT NULL DEFAULT '',
+  `type_alias` varchar(255) NOT NULL DEFAULT '',
+  `title_holder` varchar(255),
+  `table_values` varchar(255),
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 DEFAULT COLLATE=utf8mb4_unicode_ci;
+
+INSERT INTO `#__user_logs_tables_data` (`id`, `type_title`, `type_alias`, `title_holder`, `table_values`) VALUES
```
joomdonation: Could you please explain the reason of storing table_type and table_prefix instead of storing table name (like #__content) and primary key (like id)

From what I see, this data is only used in **getDataByPks** primary method. It has some problem:

1. It requires table class to be auto-loaded which is not always happen (For example, JTable::getInstance('Banner', 'BannersTable'); will return false because BannersTableBanner class is not auto-loaded, so the code in getDataByPks won't work for banners

2. Storing table name and primary key will also allow us to get data of all records in a single query instead of multiple queries, so performance would be better

For reference, here is the code which is currently use this data:
```php
public static function getDataByPks($pks, $field, $tableType, $tablePrefix = 'JTable')
{
	$items = array();
	$table = JTable::getInstance($tableType, $tablePrefix);

	if ($table === false)
	{
		return $items;
	}

	foreach ($pks as $pk)
	{
		if ($table->load($pk))
		{
			$items[] = $table->get($field);
		}
	}

	return $items;
}
```

joomdonation: Ping @mbabker  and @alikon

mbabker: Use the `#__content_types` table as an example of why this is OK given our environment.  In the PHP code you can make some assumptions about how to find `JTable` classes for components following the core MVC structure.

mbabker: That said I also haven't spent a lot of time in this code yet, so I don't really know the capabilities of it or why the data structure is what it is.  Just commenting strictly on the "why do it this way" question specifically.

joomdonation: So I guess before calling getDataByPks, we will have to call JTable::addIncludePath() method to make sure the table class is accessible

Not sure if it's used for any other purpose but for this one getDataByPks, I think it's not good as storing table name and primary column directly

mbabker: Depends what is being logged I guess.

I use a revision logging system on a couple of Laravel and Symfony projects, the tables for that revision logging includes two columns; one for the model class name (our `JTable` object basically) and one for the PK of that record.  So it's not that far off base to store references to PHP classes here, but again a lot of it is going to be highly dependent on what actually is happening.

joomdonation: Right now, it's only used to get title of items which are being published/unpublished. I don't see it's used for anything else

alikon:
not sure i can tell you the reason for that choice, i weren't in that gsoc team, 
if you submit a pr should  be better to evaluate the pros/cons :smile:

joomdonation:
@brianteeman You asked for Purge button. What happens when that button is clicked? Let me know so that we can get it implemented

installation/sql/mysql/joomla.sql
```diff
@@ -1403,7 +1406,8 @@ INSERT INTO `#__menu` (`id`, `menutype`, `title`, `alias`, `note`, `path`, `link
...
+(24, 'main', 'com_userlogs', 'User Logs', '', 'User logs', 'index.php?option=com_userlogs', 'component', 1, 1, 1, 36, 0, '0000-00-00 00:00:00', 0, 0, 'class:privacy', 0, '', 43, 44, 0, '*', 1),
```
sandewt: > Missing link in components menu to administrator/index.php?option=com_privacy yet. 

See my comment issue #15.

The same problem occurs here. Missing te menu link to `administrator/index.php?option=com_userlogs`

Looking at the code, I see there are some spaces at the end of rule 1409.

```
(24, 'main', 'com_userlogs', 'User Logs', '', 'User logs', 'index.php?option=com_userlogs', 'component', 1, 1, 1, 36, 0, '0000-00-00 00:00:00', 0, 0, 'class:privacy', 0, '', 43, 44, 0, '*', 1), 

```
Probably this has nothing to do with the missing menu links. But who knows.

sandewt:
Extra information: missing menu link to User Logs. 

menulinks

joomdonation:
@brianteeman Thanks. Added that button to my PR

plugins/system/userlogs/userlogs.xml
```diff
@@ -0,0 +1,56 @@
...
+				<field
+					name="loggable_extensions"
+					type="logtype"
+					label="PLG_SYSTEM_USERLOGS_LOG_EXTENSIONS"
+					description="PLG_SYSTEM_USERLOGS_LOG_EXTENSIONS_DESC"
+					multiple="true"
+					default="com_banners,com_cache,com_categories,com_config,com_contact,com_content,com_installer,com_media,com_menus,com_messages,com_modules,com_newsfeeds,com_plugins,com_redirect,com_tags,com_templates,com_users"
+				/>
+				<field
+						name="lastrun"
```
sandewt: Rules 48 - ...
Please remove trailing tabs.

alikon:
ok folks, code has changed "quite a bit" in the meantime :smile: thanks to the huge work of @joomdonation, we are 75% features complete see #31 for the left 25%
may i ask you some test/review/feedback ?

ot2sen:
Getting a "×
Error
Table 'privacyframe.#__user_logs' doesn't exist Table 'privacyframe.#__user_logs' doesn't exist" when having discovered and installed the two extensions.

alikon:
sounds like you need to hit the fix database button

installation/sql/mysql/joomla.sql
joomdonation: @alikon Could you please look at this issue? The menu item is missing on new installation as @sandewt  reported

ot2sen:
Yes. Did the trick. Component and plugin now shown.
I am not able to select any events in the plugin which makes it hard to test logging. Is that a known part of the 25% left?
18-05-2018_user_logs_selectable_events_missing

joomdonation:
@ot2sen That's not 25% left. Maybe you can install the branch directly https://github.com/joomla-projects/privacy-framework/archive/feature/logging.zip for testing for now?

The parameters look OK for me on new installation

logs

joomdonation:
Since the menu item is not added to backend, during testing, if you want to see list of logged actions, you will need to type the URL of the component directly http://localhost/cms/administrator/index.php?option=com_userlogs

ot2sen:
@alikon Now that the settings for events moved from plugin to component settings I still have the same issue of being unable to select events to log.

19-05-2018_useractionlogs_com_settings_missing_fields

joomdonation:
@ot2sen Is this a new installation? Or you are using patch tester to apply the patch? I made a new installation about one hour ago and it is OK. Maybe you can check to see if there are data in #__user_logs_extensions table in your installation?

alikon:
with a fresh install unable to reproduce too

ot2sen:
Yes, applied with patch tester, discover, install.

alikon:
can you test downloading this branch instaed  ?

ot2sen:
Ok, the events, which I don´t have with patch, are listed in a clean install of branch. As joomdonation mention, the admin menu item is missing, but that one I do have using the patch. So something works here and something works there. All in all it looks good if the two are combined clean branch+PR.

19-05-2018_userlogs_branck_events_ok
19-05-2018_userlogs_loggingfine

alikon:
quite confused what is still missed on a fresh branch install ?

joomdonation:
@alikon We are missing menu item to link to User Logs component with new installation

joomdonation:
Actually, just tried again and see that we are only missing the menu item when install with sample data. Without installing sample data, the menu item is there. So all we need is update sample data sql with our two new components (com_privacy and com_userlogs).

mbabker:
After merging @roland-d new fields plugin, and deconflicting that with the consent plugin, you need to use 483 and 484 for your plugin IDs now.

mbabker:
For updates you need to add some stuff to the `com_admin` script.php file similar to [those for `com_privacy`](https://github.com/joomla-projects/privacy-framework/commit/4b0f1c1bbb314bf6d9377f8187f9549d4158de9f#diff-e3c745afdbb43914c4f183a8c1587346).

administrator/components/com_userlogs/helpers/userlogs.php
```diff
@@ -0,0 +1,217 @@
...
+
+			fputcsv($fp, $log, $csvDelimiter);
+		}
+
+		rewind($fp);
+		$content = stream_get_contents($fp);
+		echo $content;
+		fclose($fp);
+
+		$app->close();
```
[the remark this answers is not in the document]
joomdonation: Yes. I wanted to ask about it.  Will make a PR for that.

libraries/src/Extension/ExtensionHelper.php
```diff
@@ -228,6 +228,7 @@ class ExtensionHelper
 		array('plugin', 'sessiongc', 'system', 0),
 		array('plugin', 'stats', 'system', 0),
 		array('plugin', 'updatenotification', 'system', 0),
+		array('plugin', 'userlogs', 'system', 0),
```
mbabker: The component and other plugin need to be added here.

alikon: done

alikon:
plugin id's are now 483, 484 (https://github.com/joomla/joomla-cms/issues/20500 docet)

joomdonation:
I also have some concerns about the terms we are using related to this feature:

1. The component name is **com_userlogs**. Actually, we are logging user's actions here, so I am not sure if this's a good name. Maybe **com_actionlog** could be a better name?

2. The same question is name of our system plugin. It's **PlgSystemUserLogs** at the moment. Is **PlgSystemActionLog** is a better name?

3. The plugin group is **actionlogs** at the moment. Should it be **actionlogs** or **actionlog** or if we decided to use **com_userlogs** as name of component, should plugin group also be renamed to **userlogs** or **userlog** to be consistent?

joomdonation:
We should also review some language items use in the extension:

1. In component config screen, we have this parameter **Select events to be logged**. Should it be changed to **Select extensions to be logged** because we allow admin to choose which extensions to log user's actions here. The word events seem not correct to me

2. Same question for account edit screen. Should **Select events to be notified for** be changed to **Select extensions to be notified for**?

mbabker:
As far as semantic naming goes it does seem action logs would be better.

As for extension names, component names tend to be more plural and plugin groups tend to be more singular.  So I'd do `com_actionlogs` and `actionlog` for naming myself.

As for storing in the database versus using plugin events, there really isn't a "better" option between database or plugin events.  I'm using plugins to collect data in `com_privacy` because it's a low frequency action and IMO that doesn't need to be persisted in long term storage (the database).  I still don't know all the specifics of this stuff yet (it's on my list for this week), but my gut instinct says we don't need database tables to know what extensions are loggable through an integration here or what content types that covers, that can be dynamically collected through plugin events as needed (event dispatching really isn't an expensive operation).

joomdonation:
Thanks for the feedback @mbabker. About the component name and plugin group, I agree that it's good names, so I will work to rename it

About using plugin events to get loggable components and the content types it supports, I just think use plugin events will be more clear and easier for third party developers to integrate it. Most of the time, each extension will only has one loggable component and few content types, so return it in plugin events seems easier than checking and inserting into database during installation/upgrade process (we are using that method in our search plugins with **onContentSearchAreas** method, and as you said, in your com_privacy, too)

alikon:
@joomdonation if you think guys,  that the plugin approach is better than the database one i'm ok too

alikon:
@joomdonation are you planning to switch  to

> using plugin events to get loggable components and the content types it supports

?

joomdonation:
@alikon As we are unsure which one is good, I think we can leave it as how it's for now. If it's needed (maybe after @mbabker review it, we can always alter it later)

mbabker:
If that's the only roadblock to merging this, I'd say keep it as is for now and as we work through to a beta release we can decide if it makes sense to change it.

joomdonation:
@mbabker Yes, that's the only thing left for now. The feature was completed and ready for human testing.

administrator/modules/mod_latestactions/helper.php
```diff
@@ -0,0 +1,49 @@
...
+ * @copyright   Copyright (C) 2005 - 2018 Open Source Matters, Inc. All rights reserved.
+ * @license     GNU General Public License version 2 or later; see LICENSE.txt
+ */
+
+defined('_JEXEC') or die;
+
+/**
+ * Helper for mod_latestactions
+ *
+ * @since  3.9.0
```
manojLondhe: Should be `__DEPLOY_VERSION__`

administrator/modules/mod_latestactions/mod_latestactions.xml
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="utf-8"?>
+<extension type="module" version="3.1" client="administrator" method="upgrade">
+	<name>mod_latestactions</name>
+	<author>Joomla! Project</author>
+	<creationDate>May 2018</creationDate>
+	<copyright>Copyright (C) 2005 - 2018 Open Source Matters. All rights reserved.</copyright>
+	<license>GNU General Public License version 2 or later; see LICENSE.txt</license>
+	<authorEmail>admin@joomla.org</authorEmail>
+	<authorUrl>www.joomla.org</authorUrl>
+	<version>3.9.0</version>
```
manojLondhe: This should be `__DEPLOY_VERSION__`

brianteeman: In this case it is fine because we know for certain it will be in 3.9 - bit of an exception

manojLondhe:
I tried several times to test this but I am not seeing anything in config options to tick mark for enabling logging for

Same isse as described here https://github.com/joomla-projects/privacy-framework/pull/20#issuecomment-390392000

Steps I took `Cloned repo -> checked out this branch -> fixed db -> discover and install -> enabled plugins`

administrator/modules/mod_latestactions/mod_latestactions.xml
manojLondhe: :+1:

alikon:
maybe we are missing to store params in .sql install, for now just save param before to go beyond ... :face_with_head_bandage:

mbabker:
Thanks to all involved here (and those from GSoC not in this effort), merged to main branch.  We can keep iterating and improving as needed but we're at the point now where it's complete enough to build upon.

joomdonation:
Great. Thanks for merging .

PR #22, opened 2018-05-07
Move view logs permission check from view to controller

joomdonation:
Pull Request for Issue # .

### Summary of Changes
As title say, this PR remove **View Logs** permission check from View class to Controller class.


brianteeman:
This code was specifically added to address https://github.com/joomla-projects/privacy-framework/issues/11

alikon:
removing is easy...
still not convinced 100% about it https://github.com/joomla-projects/privacy-framework/pull/20#issuecomment-386971749
it would be more usefull to fix pont 2.... at least for now....

joomdonation:
I still don't understand why we need this extra permission setting when we only have one view in the component. Users who has **Access Administration Interface** permission should be allowed to access to this **User Actions Log** function. But I will leave it to you for making decision

joomdonation:
OK, so we will leave the permission settings as how it is.  I updated the PR to move code which check permission from View class to display method of controller.

PR #24, opened 2018-05-08
Clean up delete logs code

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR makes several clean up to the delete logs code:

- Move delete permission check to controller
- Re-use delete method from AdminController class
- Write query to delete records directly instead of creating table object and call it's delete method.


administrator/components/com_userlogs/models/userlogs.php
```diff
@@ -268,29 +268,23 @@ public function getLogsData($pks = null)
...
 	{
-		// Check for request forgeries
-		JSession::checkToken() or die(JText::_('JINVALID_TOKEN'));
-
-		// Get the table
-		$table = $this->getTable('Userlogs', 'JTable');
+		$db    = $this->getDbo();
+		$query = $db->getQuery(true)
+			->delete('#__user_logs')
+			->where('id IN (' . implode(',', $pks) . ')');
```
alikon: please use `quoteName`

joomdonation: Yes, done.

PR #25, opened 2018-05-08
Clean up UserlogsHelper

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR makes several clean up to UserlogsHelper class

1. The code **$data = json_decode(json_encode($data), true);** in dataToCsv method looks strange to me.  End code data in json, then decode it immediately to get return data in array could cause performance issue (just guess). I removed that code and add some code to build csv data in for loop block

2. Use $db->loadObject instead of $db->loadObjectList() in getLogMessageParams

3. Added a check to prevent fatal error when $table class not found in getDataByPks method

### DIFF
--- a/administrator/components/com_userlogs/helpers/userlogs.php
+++ b/administrator/components/com_userlogs/helpers/userlogs.php
@@ -19,7 +19,7 @@ class UserlogsHelper
 	/**
 	 * Method to extract data array of objects into CSV file
 	 *
-	 * @param   array  $data  Has the data to be exported
+	 * @param   array $data The logs data to be exported
 	 *
 	 * @return  void
 	 *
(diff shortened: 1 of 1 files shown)


administrator/components/com_userlogs/helpers/userlogs.php
```diff
@@ -45,10 +44,16 @@ public static function dataToCsv($data)
...
-			$log['ip_address'] = JText::_($log['ip_address']);
+			$log               = array();
+			$log['id']         = $row->id;
+			$log['message']    = $row->message;
+			$log['log_date']   = $row->log_date;
+			$log['extension']  = $row->extension;
+			$log['user_id']    = $row->user_id;
+			$log['ip_address'] = JText::_($row->ip_address);
+			$app->triggerEvent('onLogMessagePrepare', array(&$log['message'], $log['extension']));
 			$log['extension'] = self::translateExtensionName(strtoupper(strtok($log['extension'], '.')));
```
alikon: `$log['extension']` assignment should be run before of line 56
`$app->triggerEvent('onLogMessagePrepare', array(&$log['message'], $log['extension']));`

joomdonation: The original code was working like that, I just did some clean up. Anyway, I made the change as requested.

PR #26, opened 2018-05-09
Clean up  system userlogs plugin

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR makes some clean up to system userlogs plugin. 
- Correct return type of methods and data type of some method parameters in docblock
- Remove some un-used variables
- Fix some typos
- Remove some necessary lines of code (no need for calling $mailer->setSender($sender); because it is already processed in JFactory::getMailer();

### DIFF
--- a/plugins/system/userlogs/userlogs.php
+++ b/plugins/system/userlogs/userlogs.php
@@ -120,9 +120,9 @@ class PlgSystemUserLogs extends JPlugin
 		}
 		catch (RuntimeException $e)
 		{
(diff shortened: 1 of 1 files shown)


joomdonation:
Am I correct that two settings **Send notifications for User Actions Log** and **Select events to be notified for** are only available on user profile (from administrator and frontend) page of the users has **View Logs** permission?

If I am wrong, please let me know so that the second commit in this PR need to be reverted

PR #27, opened 2018-05-10
Move delete logs process to system plugin

joomdonation:
Pull Request for Issue # .

### Summary of Changes
As mentioned, I don't think delete old log records on display process is correct. So in this PR, I move that delete process to the system plugin

Currently, it is hard coded to be run once per day (this logic is borrowed from update notification plugin). Maybe we should introduce a new parameter for configuring it? If Yes, could you recommend the title for that parameter (like Delete Frequency)?


alikon:
i wouldn't add that new parameter, at least for now, once a day can be enough

PR #33, opened 2018-05-13
[Feature] Frontend Information Request Processing

mbabker:
This will integrate the frontend for `com_privacy` to be able to submit and confirm information requests.  Functionality added:

- The email sent from the backend when an admin generates a request now includes the confirmation page's link
- A frontend confirm view which has the form where the user will enter their email address and confirmation token
- Controller task for handling the validation
    - Form requires email and token
    - Model validates the user information and the confirmation is within the 24 hour expiry
    - If successful the request is transitioned to confirmed, the user is redirected to the site homepage
- A frontend request view which is used for creating new information requests
    - For now this will be restricted to authenticated users, we can always revisit opening this up later but that should ensure the form can have captcha hooked up otherwise the submit action is a potential spam point
- Controller task for handling request creation
    - Form requires email and request type
    - If submitter is authenticated, model will associate to current user
    - Model checks for an active request and alerts the user there is already one open and to contact the site owner for an update
- Menu items for these two forms


peteruoi:
https://takeout.google.com
Just a peek at an optimal sollution(funded by the no1 company ofc)?

mbabker:
We may expand this to allow you to pick and choose what data you want, but that's far outside scope for what is going to be the deliverables to get this merged to core.

mbabker:
For what this is it's feature complete now.  Needs a lot of UX love but it's working.

ot2sen:
@mbabker All though this is closed I hope comments get heard.
The Create request  menu link in front has a UI issue that looks wrong for the purpose. Having the 'Remember me' tick box and text at this page isn´t what one expect when doing the request.
17-05-2018_createrequest_rememberme

mbabker:
You have to log in before you get to the request form.  The request form doesn't have username/password when you get there.

ot2sen:
Yeah, created is a public link :) Playing my stupid as best. Get the Export/Remove when logged. Fair enough, admin just has to do their task properly. Testing on

mbabker:
I did say people who are better with UI/UX matters than me are going to have to come behind me and change things from "functional" to "pretty" 🤣

brianteeman:
Sorry but with traveling from event to event and trying to survive in
between I have not had time

mbabker:
It's all good.  I'd rather us be in a state where the functionality is there and we work on polishing it all up than be sitting around waiting for things to be polished up and pretty before moving on to another task.  Otherwise things will never get done.

Issue #31, opened 2018-05-12
[feature/logging] How to support third party plugins?


joomdonation:
We need to find out how to allow loggin actions from third party extensions. Something come up in my mind:

1.  If extension just uses core event triggers like **onContentAfterSave**, **onContentAfterDelete**, then they will only has to insert some records related to the extension into two tables: **#__user_logs_extensions**,  **#__user_logs_tables_data**. The only thing we have to do is decide where the log related language items from that extension is stored so that we can load it to translate the message

2. If extension uses custom event trigger (for example, when someone makes an order in a shopping cart extension), they will have to write custom plugin to save data into #__user_logs table. Some questions related to this:

- Should we introduce a new plugin group for these plugins like userlogs or actionlogs? If we introduce a new plugin group, we will have to import this plugin group somewhere in our current system user logs plugin. Also, in this case, should we move our current logs method to a separate plugin like joomla and move it into this new plugin group, too?
- How do we allow these plugins re-use our addLogsToDb method (basically, store data into database and send notification emails to users choose to receive log messages). Move these methods to component model?

mbabker:
Go the plugin route and build some support infrastructure to make that easier (similar to Smart Search or Custom Fields plugins).  While it means there are more extensions in the system, we can't at a core level really try to do everything on behalf of everyone, the best we can do it make it easy to hook into the system and have things just work.

joomdonation:
Can we introduce a new plugin group like actionlogs (same as fields, search, finder...)? Then related to user action logs, we will have:

1. A system plugin (as how we are having at the moment)
2. The code which record action logs related to Joomla core will be moved into a new plugin, we can call it joomla or core and put it inside that new plugin group
3. Thid party extensions can add new action logs plugin for their extension and have these plugins inside actionlogs group, too

Is that the right direction?

mbabker:
> Is that the right direction?

Yes.

joomdonation:
Great, thanks. That's all I need to know now.

joomdonation:
Added PR https://github.com/joomla-projects/privacy-framework/pull/38

PR #34, opened 2018-05-14
Improvements

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR makes massive changes to current User Action logs code base. It contains following improvements:
- Added link to the user account performs the action and link to the item being changed by the action
- Store action related data in better format so that it can be used to generate the log message.
- Move the code which translates the json-encoded message to human readable message to a helper method instead of in a plugin trigger like before.
- Addressed an issue with the current implement: When we change state of multiple items, multiple records will be logged(each record for one item) but only one email sent.

There are still work left to get this PR completed, however, I want to submit it here to see whether we want to go with this direction or not before working further on it

- Update installation SQL 
- Update update SQL
- Added missing language items for other content types

action_logs

As this makes change to SQL, if someone wants to play with it, please get and install a copy of my branch https://github.com/joomdonation/privacy-framework/archive/improvements.zip

### DIFF
--- a/administrator/components/com_userlogs/helpers/userlogs.php
+++ b/administrator/components/com_userlogs/helpers/userlogs.php
@@ -19,7 +19,7 @@ class UserlogsHelper
 	/**
 	 * Method to extract data array of objects into CSV file
 	 *
-	 * @param   array $data The logs data to be exported
+	 * @param   array  $data  The logs data to be exported
 	 *
 	 * @return  void
 	 *
@@ -86,13 +86,13 @@ class UserlogsHelper
 	/**
 	 * Get parameters to be
 	 *
-	 * @param   string   $context  The context of the content
+	 * @param   string  $context  The context of the content
 	 *
-	 * @return  mixed  An object contain type parameters, or null if not found
+	 * @return  mixed  An object contains content type parameters, or null if not found
 	 *
 	 * @since   __DEPLOY_VERSION__
 	 */
-	public static function getLogMessageParams($context)
+	public static function getLogContentTypeParams($context)
 	{
 		$db = JFactory::getDbo();
 		$query = $db->getQuery(true)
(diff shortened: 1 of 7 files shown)


brianteeman:
if i understand this correctly in order for the nice strings then they need to be added for all components into the en-GB.plg_system_userlogs.ini file.

That would seem to require a lot of duplication to me AND how would an extension be able to add the required strings

joomdonation:
Yes. Basically, for each type of object (article, category....), we will require 6 languages items (look at articles as a sample)

A custom extension, in my mind, should implement a custom plugin with it's own language file. Unfortunately, I could not find a simpler way to support this improvements without requiring extra language items.

brianteeman:
> Addressed an issue with the current implement: When we change state of multiple items, multiple records will be logged(each record for one item) but only one email sent.

This was by design - you would not want to receive 100 emails

joomdonation:
Sorry, I was not clear on this. Before making this change, if you publish 100 items, there will be only one record logged and only one email. As you mentioned somewhere, we should have 100 records but only one email and this change allow it works exactly like that

brianteeman:
> A custom extension, in my mind, should implement a custom plugin with it's own language file. Unfortunately, I could not find a simpler way to support this improvements without requiring extra language items.

is that even possible? and if it is then it seems a very inefficient option.

alikon:
i've played a little bit with you branch  and it seems to me that with this pr we are moving in a good direction, surely better than before.
So i invite you to go on and complete this pr, so we can move togheter with the next natural step 
that you have already pointed out https://github.com/joomla-projects/privacy-framework/issues/31

joomdonation:
Thanks. One problem with this PR is it requires more language items than it should and it will require more time for translators to do the translation to different languages. I am thinking about adding some code to component helper class to get link to resources/content types (articles, categories...) and with that approach, we can reduce the language items required. I will post the post the code when I have it ready (guess I will have time to work on this tonight my time, will post it here when it's ready so that you can have a look)

alikon:
any enhancement is warmly welcomed...translators will be grateful to you

joomdonation:
OK. See https://github.com/joomla-projects/privacy-framework/pull/36/ for alternative

joomdonation:
OK @alikon. We will go with https://github.com/joomla-projects/privacy-framework/pull/36 since it's require less translation (and can avoid mistakes from translator). I'm closing this one and continue working on https://github.com/joomla-projects/privacy-framework/pull/36

PR #36, opened 2018-05-15
Improve logging

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This is an alternative for https://github.com/joomla-projects/privacy-framework/pull/34, see the original more details. 

Compare to the original PR, this reduces the number of language items for each logged content type, for now, each content type only need to have one language item for it own (to define it's type like PLG_SYSTEM_USERLOGS_TYPE_ARTICLE="article" ). However, it will require adding code to helper class of each component to get link to edit page the content type (see getContentTypeLink in ContentHelper as an example). If we want to go with this direction, I will add necessary code.


joomdonation:
OK. There might still some other clean up needed but I think you can have another look at this change now @alikon

alikon:
after a quick test  & code review is ok 4 me ..... so i'll let you finalize ... and  then i'll merge it ...

IMHO we are 1 issue left only  #31
thank you so much again @joomdonation

joomdonation:
OK @alikon. It's ready for final review and merge so that we can move to next step

PR #42, opened 2018-05-20
[feature/logging] Latest action logs module

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR implement latest action logs module to displays latest user's actions logged in our system. It's base on latest articles module. There are still something I need help to have it finished:

1. Installation sql script. @alikon Could you please help with this one? Guess we need to insert one record into #__modules and one record into #__assets table.

2. Improve module output (if needed)

module_output


mbabker:
> Guess we need to insert one record into #__modules and one record into #__assets table.

For updates if we're going to automatically display the module on the control panel then we should do this in the `com_admin` script.php file the same way component menu items and asset records are handled.

If we want it to show on new installs, then yes add the records to the `joomla.sql` files and make sure you keep the nested tree structure for the `#__assets` table in place.

administrator/components/com_admin/sql/updates/mysql/3.9.0-2018-05-20.sql
```diff
@@ -0,0 +1,4 @@
+INSERT INTO "#__extensions"
+("extension_id", "package_id", "name", "type", "element", "folder", "client_id", "enabled", "access", "protected", "manifest_cache", "params", "custom_data", "system_data", "checked_out", "checked_out_time", "ordering", "state")
+VALUES
+  (319, 0, 'mod_latestactions', 'module', 'mod_latestactions', '', 1, 0, 1, 0, '', '{}', '', '', 0, '1970-01-01 00:00:00', 0, 0);
```
alikon: for mysql null date should be 000-00-00 00:00:00

alikon:
i'm merging it despite of myself comment we can fix it in the main branch

PR #12, opened 2018-05-03
Port my plugin plg_content_consentbox as plg_content_confirmconsent for the joomla core Fixes #6

zero-24:
Pull Request for Issue #6

### Summary of Changes

Port the plugin plg_content_consentbox as plg_content_consentboxes for the joomla core (https://github.com/zero-24/plg_content_consentbox)

### Testing Instructions

apply this patch
discover install this plugin
enable the plugin
override the strings via the override manager
- `PLG_CONTENT_CONSENTBOXES_CONSENTBOX_TEXT` => Checkbox Text
- `PLG_CONTENT_CONSENTBOXES_CONSENT_MESSAGE_TEXT` => Email message

### Expected result

Required checkbox in the contact form.

image

### Actual result

No consent plugin this is a new feature

### Documentation Changes Required

Doc needs to be prepared when the consent is here to merge this. But should be pretty simple.

### Requires

To fully support com_mailto this need merged https://github.com/joomla/joomla-cms/pull/20265

PR #48, opened 2018-05-23
privacy consent plugin

brianteeman:
quick pr of the original privacy consent plugin

(I think it is missing the install sql files but the update sql is present)


administrator/components/com_admin/sql/updates/mysql/3.9.0-2018-04-01.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO `#__extensions` (`extension_id`, `package_id`, `name`, `type`, `element`, `folder`, `client_id`, `enabled`, `access`, `protected`, `manifest_cache`, `params`, `custom_data`, `system_data`, `checked_out`, `checked_out_time`, `ordering`, `state`) VALUES
+(481, 0, 'plg_user_privacyconsent', 'plugin', 'privacyconsent', 'user', 0, 0, 1, 0, '', '', '', '', 0, '0000-00-00 00:00:00', 0, 0);
```
alikon: 481 should became 485, if/when #20 will be merged and for #12 already merged

joomdonation:
@brianteeman Am I correct that user will only have to accept privacy policy in two places:

- User registration form
- User profile page

I ask this because in your code, I see you check $formName with value **com_admin.profile** (edit user profile page in administrator profile for login)

Also,  once he agreed, next time he access to his profile page, the accept privacy policy section won't be displayed anymore?

infograf768:
in https://github.com/joomla/joomla-cms/pull/20051 main issue was

> Once plugin is enabled and a user logs in, the user is indeed presented with her/his edit profile page displaying the redirect message, but the already logged in user is able to totally ignore that page and switch to any other page on the site without agreeing to the Policy. This defeats imho the purpose of this Redirect as it only prevents the user from modifying her/his profile.

brianteeman:
yes that has already been stated n the related issue - no need to repeat it

joomdonation:
So I started working on it on my own branch https://github.com/joomdonation/privacy-framework/tree/privacy_consent. See https://github.com/joomla-projects/privacy-framework/compare/dev/privacy...joomdonation:privacy_consent for the change. If someone wants to test it, please download a copy of my branch https://github.com/joomdonation/privacy-framework/archive/privacy_consent.zip, install and test it. Main changes include:

1. Change the privacy consent from a user plugin to system plugin (so that we can force user to edit profile page to accept privacy policy before they can access to other pages)

2. When user accept privacy policy, a record will be inserted into new table (#__privacy_consent) - (not use use note and profile data to track this like in old code)

3. User will only be asked to accept it one time at the moment. So if he accepted, next time when he edits profile, he won't be asked anymore (I am not sure if it's right)

4. A new view added to com_privacy to view the consent records

5. Code clean up

There are still works need to be done:

- SQL installation and update for other database types (only mysql implemented at the moment)
- Might need to review and clean up the consent view and model (I copied and modified from requests view/model). Honestly, I am not using Joomla MVC in my own extension, so I'm not really familiar with it.

joomdonation:
It's ready for testing. So please help testing it and give feedback. Once the work is done, I will need to find out how to much PR to @brianteeman branch

JoomliC:
> Change the privacy consent from a user plugin to system plugin (so that we can force user to edit profile page to accept privacy policy before they can access to other pages)

I was doing some test on my localhost yesterday, and kept the plugin as a user one, but included a new checkUserRequireConsent function in CMSApplication.
This to allow inclusion of a workflow similar to request password reset, there for Privacy consent, where it could be batched processed to require re-consent (when Privacy changes for example).

And in view that the user privacy plugin adds the user form option if enabled.

What's your opinion about this way ?

brianteeman:
1-3 - all good
4 - the filters have one extra at the end that is not needed
chrome_2018-05-24_10-49-29

joomdonation:
Thanks @brianteeman.  Made PR https://github.com/joomla-projects/privacy-framework/pull/49 , hopefully we can have your work finished this time

@JoomliC I haven't looked at how password reset is handled yet but for now, I want to have this as a system plugin. Have the code in single place is easier to maintain. Let's focus to have it works as we want now, further improvement if needed, can be done later

Issue #16, opened 2018-05-04
error.log


brianteeman:
the error log that joomla creates contains ip addresses

although it is still debatable if an ip address on its own is personally identifiable information we should err on the side of caution - we are not lawyers

so we should have a method to clear the log after a set period of time. 

it is trivial to write a plugin to delete the log file every x days

however you would probably need to keep the last x days of logs so the best way to approach it would be to have some form of log rotation and then only delete the oldest file.

the other alternative would be to only keep x number of lines (records) but that wouldnt satisfy the date requirmeent

brianteeman:
yes i think it would have to be using the log rotation method and with a plugin (dont really want to use cron unless it is essential)

Shouldnt be too hard to do that

sandewt:
I try to understand the problem, hence the following suggestion:

The `Session Data Purge plugin` uses a function, which is based on the `probability`.
May be you can also use this functionality here too, to do the log rotation check with a plugin? 
So that it will not take place with every request.

```
public function onAfterDispatch()
{
	$probability = $this->params->get('gc_probability', 1);
	$divisor     = $this->params->get('gc_divisor', 100);

	$random = $divisor * lcg_value();

	if ($probability > 0 && $random < $probability)
	{
		$this->doLogRotation();
	}
}	
```

mbabker:
For this case a log rotation plugin could run based on a timestamp (similar to the update notification plugin and how it sends emails out).  Because realistically you want log files to rotate out every X days, not every X requests (which would mean several times per day).

brianteeman:
I will work on this over the weekend

brianteeman:
If you are writing to the file every day then you have something wrong

tonypartridge:
😂 I’m not, but some people will be.... just a thought and an easy way to maintain it?

PhoenixUK:
@brianteeman 

> although it is still debatable if an ip address on its own is personally identifiable information we should err on the side of caution - we are not lawyers

Just to hopefully place a little meat on the bones above if I may (I'm not a lawyer but read on but heavily invested in Data & Cyber Security incl Privacy / GDPR for Third Sector orgs)

**1.** Firstly here is the link to a law firms PDF on various aspects surrounding Identifiability and GDPR, it is dated 2016 but is one of the best found on this subject, from an actual law firm.

[https://www.slaughterandmay.com/media/2535637/personal-data-anonymisation-and-pseudonymisation-under-the-gdpr.pdf](url)

[...]

PR #51, opened 2018-05-26
[plugin] - log files rotation

alikon:
Pull Request for Issue #16 .

### Summary of Changes

a new plugin to take care of  log files & rotation

### Testing Instructions
look at log file folder

### Expected result

log files are rotated

### Actual result

no log files rotation nor deletion


brianteeman:
I don't understand why you are clearing the cache folder.

brianteeman:
The option should be in days and not hours.

brianteeman:
Instead of listing the possible log files in an array would it not be more future proof if the plugin rotates all log files in the folder.

alikon:
> I don't understand why you are clearing the cache folder.

to have the hidden `last_run` param  udpdated

sandewt:
@alikon and @brianteeman 

At the moment I am also working on a similar plugin. :-)

[https://github.com/sandewt/plg_system_logrotation_deploy_version](https://github.com/sandewt/plg_system_logrotation_deploy_version)

alikon:
> Instead of listing the possible log files in an array would it not be more future proof if the plugin rotates all log files in the folder.

maybe yes, but i'm still concerned about the "purge" param .....

any feedback is warmly welcomed, as usual

sandewt:
@alikon 

I see that the original `error.php` file is shift to `error.php.1`. And so on. 

When I approach the URL `Joomla/administrator/logs/error.php.1` directly, the content of the log file is readable. 

Sorry, but I think this is not desirable in connection with security.

brianteeman:
yes that is correct it is not a good idea. i would rename it either 1.error.php or error.1.php

sandewt:
>Instead of listing the possible log files in an array would it not be more future proof if the plugin rotates all log files in the folder.

This should also be useful for log files which not belongs to the Joomla core.

@alikon is this easy to realize?

alikon:
@sandewt 
ok now all log files are worked, or  at least log files that match `^[a-zA-Z].*\.php$` i.e. not starting with a number as this is reserved to rotated logs

sandewt:
There are different log files e.g.:
- deprecated.php,
- error.php,
- everything.php,
- joomla_update.php,
- upload.error.php,
- jmodulehelper.log.php,
- jcontroller.log.php,
- indexer.php,

Extends the filter:
`$files = \JFolder::files($logpath, '^[a-zA-Z_\.].*\.php$', 1, false);`

joomdonation:
If I read the code right, the logic we are using here might not work well. Assume we have file indexer.php, first time the plugin run, it will be renamed to 1.indexer.php. Then the second time it runs, if the file indexer.php does not exist, the previous file 1.indexer.php won't be renamed to 2.indexer.php, so it's not working properly

Also, do we have to care about logs files located inside a sub-folders?

mbabker:
> Also, do we have to care about logs files located inside a sub-folders?

No.  Default use of the `JLog` API only writes to the directory specified in the `log_path` param (default `administrator/logs`).  Any subdirectories are going to be the result of extensions either writing their own files without using the API or having customized the path in use when they create a new logger.  Since we don't have a compiled environment, there's no sane way to deal with this (i.e. a couple extensions I've written support logging but the loggers are only registered when the component or supporting plugin are executed), so we can only deal with the configured log path and leave rotation/cleanup to the extensions doing other things.

joomdonation:
If so, I think we can simplify the code a bit more.  While reviewing the code, I see a parameter called purge  (which basically if set to Yes, will delete all the log files). Do you think it's needed?

Also, how long should we keep the logs? how many days we should run the rotate process? How many versions to rotate? That would help setting default value for these parameters and finish implementing this feature

brianteeman:
How long we keep logs for etc by default will always be an arbitrary decision. So lets go for 30 days and to keep 1 rotated log as the defaults

joomdonation:
OK. Thanks.  There is still a question about Purge parameter (so that when the plugin is run, it will delete all log files). Do you think if it's needed?

brianteeman:
re purge option - I am not sure what the intent of this option was or how it would work in practice but it doesnt sound like we need it

joomdonation:
As mentioned, when it's enabled and the plugin is run, it will empty the log folder. Seems not needed to me, too (or at least it should not be an option of this plugin)

joomdonation:
Hmm. Not sure what was wrong with the merge. In my PR, there were only some files changes https://github.com/joomla-projects/privacy-framework/pull/59/files , not sure why after merging there are so many changes to this PR :(

mbabker:
At some point somehow you reverted the merge of #20 it looks like.

joomdonation:
Maybe the easiest way to solve this would be making a new PR. @alikon Please let me know if you want me to do it.

joomdonation:
At the moment, in the code, I only check and delete the log files has version maximum to log to keeps parameter. I didn't think about the case which it has a higher value at first, then reduce the value

If we have to deal with the case, the code for deleting need to be updated

Just to be more clear, you are saying that no log files were deleted?

sandewt:
>Just to be more clear, you are saying that no log files were deleted?

**Files are being deleted.**

The test above describes the case  by changing the settings from 5 to 3 from `How many versions to keep`.

joomdonation:
So the only issue is that the files with version higher than 3 (Current value **How many versions to keep** parameter) need to be deleted, too, correct?

sandewt:
You can try some code like this.

$n = 10 is the maximum value from ` How many versions to keep for a log file`.

```
for ($n = 10; $n > $logstokeep; $n--)
{
	if (JFile::exists($path . $n . '.' . $file . '.php'))
	{
		File::delete($path . $n . '.' . $file . '.php');
	}
}
```

sandewt:
>We can have better code.

Quite right.

alikon:
> Maybe the easiest way to solve this would be making a new PR. @alikon Please let me know if you want me to do it.

 @joomdonation  yes please

alikon:
closed in favor #61

joomdonation:
Thanks @alikon

PR #55, opened 2018-05-29
[log-rotation]Improvement + bugs fixes

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This could be some improvements to the code proposed at https://github.com/joomla-projects/privacy-framework/pull/51. It changes the logic of rotating log files in the code and address some limitation from original PR:

1. Doesn't work in case there are log files in sub-folder
2. Does't rotate the files from second run when original log file was deleted (and not created before the process runs second time) (see my explanation here https://github.com/joomla-projects/privacy-framework/pull/51#issuecomment-392537244)
3. Doesn't work (at least on windows) as reported

@alikon (and @sandewt  since you are working on similar plugin), could you please check to see whether the code I propose here can be used?


joomdonation:
Do you know why I cannot access to https://ci.joomla.org/joomla-projects/privacy-framework/428 to see the drone error details? Every time I access to it, I receive **ci.joomla.org refused to connect** error. Maybe it's because I'm being blocked from accessing to it for some reasons?

brianteeman:
you cant access that link because it is not https ;)

The error from drone is

FILE: ...a-projects/privacy-framework/plugins/system/logrotation/logrotation.php
--------------------------------------------------------------------------------
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
--------------------------------------------------------------------------------
 296 | ERROR | File must end with a newline character

brianteeman:
I clicked on the Details link in the Checks info box and then on the phpcs link which was highlighted with a red cross. HOWEVER it may be that i have elevated privileges - not sure

joomdonation:
I also clicked on the details link but could not access to it. I think everyone should be allowed to see that drone error. I guess maybe my country Vietnam is being blocked by the host

sandewt:
>could you please check to see whether the code I propose here can be used?

I have made various changes due to changed insights.
Older files get a suffix (in connection with ordening).
So, you may be able to do something with the code below.

```
// Method to get a list with the original log files
private function getListOrgLogFiles()
{
	// Add the file path
	$path = $this->app->get('log_path');

	// Filter all log files back to the original file names
	$files = JFolder::files($path, '^[a-zA-Z].*\.php$', 1, false);
	$files = preg_replace('/.php|[0-9].php/', '', $files);
	
	$result = array();
	foreach ($files as $file)
	{			
		if (!in_array($file, $result))
		{
			$result[] = $file;
		}	
	}
		
	return $result;
}

// Get the list with the original log files
$files = $this->getListOrgLogFiles();

var_dump($files);
die;

// results:
// error -> error
// error1 -> error
// error2 -> error
// joomla_update1 -> joomla_update
// jmodulehelper.log1 -> jmodulehelper.log
```

joomdonation:
I guess your code will only handles the case all log files stored in the main log folder (not log files stored in sub-folders).

The idea with the code I proposed is the that code in getLogFiles method return an associative array

1. Not rotated files are stored in an array with index 0
2. The first run rotated files stored in an array with index 1
3. The second run rotated files store in an array with index 2

and so on....

Base on that, we can easily delete the files which reach the version limit (for example 5), and then rotates/increase version number if files with lower version.

It works for both log files stored in main logs folder and sub-folders (For example, akeeba subscriptions store it's log files under administrator/logs/akeebasubs_failed, not in administrator/logs folder)

sandewt:
>I guess your code will only handles the case all log files stored in the main log folder (not log files stored in sub-folders).

My code provides that only log files are stored in the main log folder.

Personal note:
I prefer that the `Log rotation check (in days)` has a greater range.
E.g. 30 days or more. (the last is now 14)
After a period of e.g. 2 updates and a frequency time of 30 days, we will always have 60 days of logs.

joomdonation:
@sandewt For default value of the parameter, it should be handled in original PR https://github.com/joomla-projects/privacy-framework/pull/51 . In this PR, I only focus solving the rotates log files problem.

To be honest, I don't know how important data in the log files, thus don't know how long it should be kept. For that, we will have to ask for advice from Michael (again on original PR).

Now I guess we can wait for @alikon to check the code to see whether he agrees with the solution propose in this PR.

sandewt:
@joomdonation thanks

>To be honest, I don't know how important data in the log files, thus don't know how long it should be kept. For that, we will have to ask for advice from Michael (again on original PR).

@brianteeman see your comment #16 
>the idea was to every 30 days

alikon:
sorry for the late response  folks, 
i agree the log rotation days param  should be handled in the main pr #51
7,14, 30 i don't care too much, let's move this on the main pr

PR #61, opened 2018-06-01
System - Log Rotation plugin

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This is a replacement for https://github.com/joomla-projects/privacy-framework/pull/51 . The original PR now could not be merged because a mistake in my PR somehow. Sorry @alikon for stealing your work.

### Testing Instructions
Install a copy of my branch, test and make sure log files are rotated and deleted properly. Ping @sandewt for human testing as he tested the original PR and pointed out a potential issue


joomdonation:
For some reasons, the plugin seems only run in backend, not frontend although it's a system plugin. Not sure why it happens. @sandewt when you test it, could you please confirm this issue?

sandewt:
>For some reasons, the plugin seems only run in backend, not frontend although it's a system plugin.

Plugin works in the backend and in the frontend.

Nevertheless, I sometimes see inexplicable behavior in the error log file.
E.g. error files that have a number that I did not expect.
At this moment I can not say any conclusions about it.
In that case, I close the browser (Firefox) and restart testing. In the hope I did not see it well.

>... and pointed out a potential issue

The pointed issue seems to be solved (How many versions to keep for a log file).

I think it is necessary to test more.

joomdonation:
> For some reasons, the plugin seems only run in backend, not frontend although it's a system plugin
Figured out. Guess I opened a different test site in the frontend and that's why I didn't see the plugin run. Tested it again now and see that the plugin is running OK on both frontend and backend

The files also rotated properly in my test, so I guess we need more tests from you and others. Thanks for help with testing.

plugins/system/logrotation/logrotation.php
```diff
@@ -0,0 +1,275 @@
...
+	 * @param   string  $filename        Name of file to rotate
+	 * @param   int     $currentVersion  The current version number
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	private function rotate($path, $filename, $currentVersion)
+	{
+		if ($currentVersion == 0)
```
alikon: use === instead

plugins/system/logrotation/logrotation.php
```diff
@@ -0,0 +1,275 @@
...
+		{
+			$rotatedFile = $path . '/1.' . $filename;
+		}
+		else
```
alikon: please clean the code to not use the else

joomdonation:
@alikon I modified code to use strict comparison as suggested. For not using else, I don't know if it's good, so I will leave it as how it's

Assume want to remove else, I guess we will have to modify code to below (not good as original code, I think).
```php
$rotatedFile = $path . '/1.' . $filename;
		
if ($currentVersion > 0)
{
	/*
	 * Rotated log file has this filename format [VERSION].[FILENAME].php. To rotate it, we just need to explode
	 * the filename into an array, increase value of first element (keep version) and implode it back to get the
	 * rotated file name
	 */
	$parts    = explode('.', $filename);
	$parts[0] = $currentVersion + 1;

	$rotatedFile = $path . '/' . implode('.', $parts);
}

File::move($path . '/' . $filename, $rotatedFile);
```

alikon:
i've tested successfully

brianteeman:
I tried to do a pr on your branch but for some reason it didn't work

Below are my suggested changed files. Note I removed the purge strings as I couldnt see them being used

**administrator\language\en-GB\en-GB.plg_system_logrotation.ini**
PLG_SYSTEM_LOGROTATION="System - Log Rotation"
PLG_SYSTEM_LOGROTATION_XML_DESCRIPTION="This plugin periodically rotates system log files."
PLG_SYSTEM_LOGROTATION_CACHETIMEOUT_LABEL="Log Rotation (in days)"
PLG_SYSTEM_LOGROTATION_CACHETIMEOUT_DESC="How often should the logs be rotated."
PLG_SYSTEM_LOGROTATION_LOGSTOKEEP_DESC="The maximum number of old logs to keep."
PLG_SYSTEM_LOGROTATION_LOGSTOKEEP_LABEL="Maximum Logs"

**administrator\language\en-GB\en-GB.plg_system_logrotation.sys.ini**
PLG_SYSTEM_LOGROTATION="System - Log Rotation"
PLG_SYSTEM_LOGROTATION_XML_DESCRIPTION="This plugin periodically rotates system log files."

joomdonation:
Thanks @brianteeman. I updated the language files as you suggested. The purge strings need to be removed as you said because that option was removed from the plugin (there was a purge parameter at the beginning but it was removed)

PR #66, opened 2018-06-04
Rename message to action

brianteeman:
Its not an email manager but an action manager


joomdonation:
Since we change "Message" to "Action", should we change the language keys as well, like:

COM_ACTIONLOGS_ACTION="Action"
COM_ACTIONLOGS_ACTION_ASC="Action ascending"
COM_ACTIONLOGS_ACTION_DESC="Action descending"

If we do that, we have to modify code to use new language keys, but that should be easy

brianteeman:
I was still thinking in the mindset that this was existing cod e so I couldn't change that. You are correct and I will make the relevant changes.

brianteeman:
Constants renamed now as well

joomdonation:
Thanks, it's good now

brianteeman:
Thanks

PR #82, opened 2018-06-13
remove unused strings

brianteeman:
These are from my original plugin that had a postinstall message. We don't need them anymore (there will be a seperate pr for a new postinstall message)


brianteeman:
Thanks

PR #83, opened 2018-06-13
User Plugin for terms and conditions

brianteeman:
Similar to the existing terms field in the user profile plugin but standalone

Only present on the registration form

No logging and no redirect for existing users


plugins/user/terms/terms.php
```diff
@@ -0,0 +1,141 @@
...
+		{
+			return true;
+		}
+
+		// Check that the terms is checked if required ie only in registration from frontend.
+		$option = $this->app->input->getCmd('option');
+		$task   = $this->app->input->get->getCmd('task');
+		$form   = $this->app->input->post->get('jform', array(), 'array');
+
+		if ($option == 'com_users' && in_array($task, array('registration.register')))
```
joomdonation: Please change this line of code to, otherwise, registration won't be processed when this plugin is enabled (the condition always return true during user registration)

```php
if ($option == 'com_users'
	&& in_array($task, array('registration.register'))
	&& empty($form['terms']['terms']))
{
	throw new InvalidArgumentException(Text::_('PLG_USER_TERMS_FIELD_ERROR'));
}
```

plugins/user/terms/field/terms.php
```diff
@@ -0,0 +1,136 @@
...
+	/**
+	 * Method to get the field label markup.
+	 *
+	 * @return  string  The field label markup.
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	protected function getLabel()
+	{
+		if ($this->hidden)
```
joomdonation: I'm not sure when this code is used? Maybe it is only needed for hidden custom field type and for this field type, it could be removed?

plugins/user/terms/terms.php
joomdonation: @brianteeman Could you please fix this error so that we can get this PR merged? I tested it and it's all working well except this error

joomdonation:
It's good now. Thanks. Could you please resolve conflicts too? After that, I think it's good to merge

brianteeman:
merge conflicts resolved

administrator/components/com_admin/sql/updates/sqlazure/3.9.0-2018-06-13.sql
```diff
@@ -1,2 +1,3 @@
 INSERT INTO "#__extensions" ("extension_id", "package_id", "name", "type", "element", "folder", "client_id", "enabled", "access", "protected", "manifest_cache", "params", "custom_data", "system_data", "checked_out", "checked_out_time", "ordering", "state") VALUES
 (488, 0, 'plg_quickicon_privacycheck', 'plugin', 'privacycheck', 'quickicon', 0, 1, 1, 0, '', '', '', '', 0, '1900-01-01 00:00:00', 0, 0);
```
joomdonation: Guess this should have ID = 489 like others?

brianteeman: oops

joomdonation:
It's all good now, thanks :)

brianteeman:
Thanks

PR #85, opened 2018-06-14
[Feature] - Remove request

alikon:
### Summary of Changes
added basic code from https://github.com/joomla/joomla-cms/pull/20173 for the remove request 
added two events for core & 3dp

- `onPrivacyCanRemoveData`
For core it check for block remove if the user is a Super Admin  
and listen from  3dp event for a reason to block the remove request

- `onPrivacyRemoveData`
For core it pseudo anonymize the #__user table fields (name,username,email)
trigger the event for 3dp

it logs any remove request


mbabker:
Similar to export, this should be done through plugins and not us just arbitrarily changing the user account here.  Personally, I intended to have two plugin events on the removal path before my time got deflected elsewhere:

- `onPrivacyCanRemoveData($request)` (sanely covered by your before event)
- `onPrivacyRemoveData($request)`

The can remove event is important because the system should have a way for someone to say "whoa, we can't remove this account data, and here's why" and block the removal (this should primarily be cases where the entire user account should be untouched, you probably wouldn't want the system to service a removal request for a super user account without additional manual intervention first).

alikon:
ok now the `onPrivacyCanRemoveData` plugin event responses are checked before removal...
For core it check for not remove a Super Admin as suggested

do you think we should care of `onPrivacyRemoveData` too ?  
imo no....

mbabker:
OK I mildly hijacked things a little bit here, but, things should be "ready" now and consistent with the other stuff I've done.  Things I did:

- The check if a data can be removed uses a data object, similar to the export process using them (OK, I've got a think for using objects with known fields versus unkeyed arrays)
- Moved user anonymization into the user plugin and adjusted the data used for anonymized accounts
- Filled in the language strings
- Added log action for an export being blocked by a plugin
- Removal requests should always have the event dispatched, even if a user account isn't associated with the request (imagine a commenting platform where users aren't required to have accounts to comment, a removal request should still be able to update that "anonymous" entry)

alikon:
ready to go for me....

btw, the last point is a bit obscure to me, what we can anonymize from an already

>  "anonymous" entry

?

mbabker:
The removal process is intended to anonymize/remove/whatever the data it is told to, not only for registered users.  In the case of someone commenting on a blog when giving only their name and email address (and they don't have an account on the site, so they are essentially "anonymous" in the sense there isn't a known user account), a plugin can handle anonymizing the user info shown on the comment, unpublishing the comment, deleting it, or whatever the case may be.

alikon:
crystal clear now

Issue #72, opened 2018-06-06
[Privacy] Admin Notifications


mbabker:
There is no form of notification to admins when requests are created.  If we get the stuff in #5 done that might at least give them a spot to look at, but we probably should add some form of notifications making use of the messaging component or emails or something.

brianteeman:
Now that the messaging comment is usable I would probably use that

mbabker:
PR - #88

PR #88, opened 2018-06-15
[Feature] Admin Notifications

mbabker:
Pull Request for Issue #72

### Summary of Changes

Adds notifications in the backend via com_messages when a user creates or confirms their information requests.  Also adds a helper method to `MessagesModelMessage` to send a message to all super users.

### Testing Instructions

Create an information request from the frontend, and confirm an information request created via either the backend or frontend.  The frontend submission and the confirmation should add messages to the message component.


administrator/components/com_messages/models/message.php
```diff
@@ -409,4 +409,95 @@ public function save($data)
...
+			{
+				$this->setError(JText::_('COM_MESSAGES_ERROR_NO_USERS_SET_AS_SUPER_USER'));
+
+				return false;
+			}
+
+			foreach ($userIDs as $id)
+			{
+				/*
+				 * All messages must have a valid from user, we have use cases where an unauthenticated user may trigger this
```
brianteeman: shouldnt the from user be the one configured in global configuration

mbabker: The from user in com_messages is a Joomla user (since the component's designed to be an internal messaging thing from the looks of the data model).  The one in global config is specific to email.

brianteeman: I was thinking these were emails ;(

mbabker: com_messages also sends an email if the user is configured to receive them.

mbabker: It's the same as the new article notification basically.

brianteeman: The new Article notification email comes from the email address defined in the global configuration. If it didnt then it would probably be get blocked as spam

mbabker: Right.

In com_messages, the from user is defined as the Joomla user the message is from.  When you view the message in the Joomla backend, that is who is shown as the sender.

When com_messages sends the email out, it uses the email configuration for your site, so the global config's sender (from address) is used there, and the email's reply-to address is set as the Joomla user account's email.

Your confusion here might be the naming of the parameter.  All this is doing is defining the Joomla user account the message originated from since com_messages is designed as an internal message platform.  It doesn't impact the sender information on the actual email that is sent (if configured to do so).

mbabker: What the comment here is indicating is that we're setting a value so [this block](https://github.com/joomla-projects/privacy-framework/blob/dev/privacy/administrator/components/com_messages/models/message.php#L304-L308) doesn't get triggered and a potentially null value be set, since the confirm action can be done without being authenticated.  When you get to the [sending the email](https://github.com/joomla-projects/privacy-framework/blob/dev/privacy/administrator/components/com_messages/models/message.php#L360-L407) part that "from user" is only used to set a reply address to the email that is actually sent, it doesn't change the From address on the email signature, so that uses the global config data correctly.

brianteeman: Ok great. Best to be safe and check

PR #91, opened 2018-06-15
Privacy quick icon clean up

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR makes some clean up to privacy quick icon:

1. Change logic of the original getOlder method. The original code looks strange to me, all it needs is return number of requests need admin attention. There was strange lines of code to create an array and it looks wrong to me

2. Use JResponseJson to return the ajax data, thus the javascript code need to be changed, too.

3. Improve the code when token is invalid. The original code doesn't look correct, the new code is copied from com_installer update controller, ajax method 

4. User better method names (unsure if it's better :D).

ajax -> get_number_urgent_requests
older -> getNumberUrgentRequests

### DIFF
--- a/administrator/components/com_privacy/controller.php
+++ b/administrator/components/com_privacy/controller.php
@@ -50,29 +50,30 @@ class PrivacyController extends JControllerLegacy
 	}
 
 	/**
-	 * Fetch and report privacy requests in JSON format, for AJAX requests
+	 * Fetch and report number urgent privacy requests in JSON format, for AJAX requests
 	 *
 	 * @return void
 	 *
 	 * @since __DEPLOY_VERSION__
 	 */
-	public function ajax()
+	public function get_number_urgent_requests()
 	{
 		$app = Factory::getApplication();
 
 		// Check for a valid token. If invalid, send a 403 with the error message.
 		if (!Session::checkToken('get'))
 		{
-			$response = new JsonResponse(new \Exception(Text::_('JINVALID_TOKEN'), 403));
+			$app->setHeader('status', 403, true);
 			$app->sendHeaders();
-			echo json_encode($response);
+			echo JText::_('JINVALID_TOKEN');
 			$app->close();
 		}
 
 		/** @var PrivacyModelRequests $model */
-		$model    = $this->getModel('requests');	
-		$requests = $model->getOlder();
-		echo json_encode($requests);
+		$model                = $this->getModel('requests');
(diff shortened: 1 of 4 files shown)


alikon:
there was a "lazy reason" for 1. "strange logic", me not  a js skilled one :smile:  :face_with_head_bandage:  so i've used  something already working (extensionupdate.js) thus for 2.

alikon:
for 3. can you elaborate a bit more , what's wrong with this snippet 

```php
		// Check for a valid token. If invalid, send a 403 with the error message.
		if (!Session::checkToken('get'))
		{
			$response = new JsonResponse(new \Exception(Text::_('JINVALID_TOKEN'), 403));
			$app->sendHeaders();
			echo json_encode($response);
			$app->close();
		}
```

alikon:
for 4. maybe
ajax -> get_urgent_requests
older -> getUrgentRequests

mbabker:
Well, AJAX method should still be camelCase to follow conventions, otherwise it's better than before 😉

joomdonation:
@alikon When I looked at the code quickly, I thought it's incorrect to pass Exception to JsonResponse constructor but checking the code again, I see that I'm wrong. There was still something with the original code:

- The response code is 200 instead of 403 as we want. To be honest, I don't know what should be the http code returned in this case 200 or 403 when token is invalid. Need advice from @mbabker 
- We should not have to call json_encode again on $response object. We can call the code directly:

echo  new JsonResponse(new \Exception(Text::_('JINVALID_TOKEN'), 403));

So if we expect to receive 403 status code, then the original code could be changed to:
```php
if (!Session::checkToken('get'))
{
	$app->setHeader('status', 403, true);
	$app->sendHeaders();
	echo  new JsonResponse(new \Exception(Text::_('JINVALID_TOKEN'), 403));
	$app->close();
}
```

If we expect to receive 200 status code, then the original code could be:
```php
if (!Session::checkToken('get'))
{	
	echo  new JsonResponse(new \Exception(Text::_('JINVALID_TOKEN'), 403));
	$app->close();
}
```
@mbabker Could you help making decision for this?

joomdonation:
For the name of methods, we actual just return number of requests, not the list of requests, so **getNumberUrgentRequests** is more correct to me. But I won't complain much if we change it to getUrgentRequests as suggested.

mbabker:
It definitely should not be a 200 response, so the 403 bit is correct.

PR #78, opened 2018-06-12
[plugin][quickicon] - notify if requests are older than X

alikon:
Pull Request for Issue #5 .

### Summary of Changes

- added a new com_privacy option parameter
- a new quickicon plugin that notify when there are privacy requests older than the number of days setted in the component option

### Expected result
screenshot from 2018-06-12 17-33-40

PR #92, opened 2018-06-16
[Feature] Privacy Dashboard

mbabker:
Pull Request for Issue #5

### Summary of Changes

Creates a new dashboard view for `com_privacy`.  Nothing extremely fancy, but functional enough to be informative.

screen shot 2018-06-15 at 10 17 32 pm

Right now it has two "modules":

- Request counts (basically the same as `mod_privacy_dashboard`)
- Status check to help guide users and ensure all resources are available or that there are no anomalies (i.e. "old" requests needing attention, kind of requires #78 to be finished first since that's setting the conditions to determine "old")

Also cleans up the markup for the actual module a bit (though to be honest either the component should just load the module or we move the markup to a layout so that's not 1-for-1 duplicated).

### Testing Instructions

Apply patch, make sure things look sane

- New plugin event to report status of privacy policy being published


mbabker:
Added urgent request count check based on #78 setup so should be good here.

joomdonation:
I tested it and it works fine.

alikon:
tested successfully

added a missed string COM_PRIVACY_DATA_REMOVED

p.s
sorry i've again hit the wrong button

PR #96, opened 2018-06-17
Fix frontend submit request error

joomdonation:
Pull Request for Issue # .

### Summary of Changes
At the moment, when we try to submit a request from frontend, there is fatal error:

> Call to a member function bind() on boolean	
> D:\www\privacy-framework\administrator\components\com_messages\models\message.php:297

This PR fixes that error. 

### Testing Instructions
1. Install latest dev/privacy branch
2. Try to create a request from frontend
3. Before patch, you will receive error above
4. After patch, the error is gone


alikon:
i've experienced the same issue and used successfully the same fix

PR #97, opened 2018-06-17
[plugin][privacyconsent] - open the privacy policy article

alikon:
### Summary of Changes
added a pseudo task

### Testing Instructions

- set a privacy policy article in the plugin privacyconsent
- create a new user
- log for the first time

click on the privacy policy article link
screenshot from 2018-06-17 10-15-47
### Expected result
 the privacy policy article open in  modal window

### Actual result
the user profile page is opened in modal window
screenshot from 2018-06-17 10-20-08


brianteeman:
I agree that there is a bug but I dont understand why

This should work the same as the TOS field in the user profile plugin and there is no task there

joomdonation:
This is honestly, not a good fix. I think we will have to change the code to something like that:
```php
$option           = $this->app->input->getCmd('option');
$task             = $this->app->input->get('task');
$view             = $this->app->input->getString('view', '');
$layout           = $this->app->input->getString('layout', '');
$id               = $this->app->input->getInt('id');
$privacyArticleId = $this->params->get('privacy_article');

// If user is already on edit profile screen or view privacy article or press update button, do nothing to avoid infinite redirect
if ($task == 'profile.save'
	|| ($option == 'com_content' && $view == 'article' && $id == $privacyArticleId)
	|| ($option == 'com_users' && $view == 'profile' && $layout == 'edit'))
{
	return;
}
```

Also, we have to care about multilingual website, too. So $privacyArticleId must be changed to the associated article in in the current language...

brianteeman:
I cannot replicate this

The link that I have in the code when this plugin is enabled is
`<a href="/privacy-framework/index.php/2-uncategorised/1-abc?tmpl=component&amp;lang=*" class="modal" rel="{handler: 'iframe', size: {x:800, y:500}}">Privacy Policy</a>`

As you can see the link is using tmpl=component so there is no way for any modules to be loaded

brianteeman:
@alikon can you still replicate this?

joomdonation:
This should still be a valid bug. You will need to login using a new account (has not consented privacy policy yet) and you should see the bug

brianteeman:
Ah in that one scenario it is a bug - the link is the same but on the edit profile page it is not using the tmpl=components

alikon:
yes without  this pr i still have the issue

joomdonation:
It's not problem with tmpl=components. It's the code which we block users from accessing to any other page before they consent privacy policy causing this bug. I will make a new PR to replace this one

brianteeman:
@joomdonation that makes sense!!!

Thanks

alikon:
why this is not a good fix ?

joomdonation:
@alikon Someone can simply insert task=privacymodal to an URL (for example, URL of an article) to bypass this check

brianteeman:
We didn't see it in core before where this same url block is used (reset password required) because the tos plugin isnt enabled on the edit profile page

brianteeman:
Closing as I merged #117 

Thanks everyone

PR #99, opened 2018-06-17
Help buttons

brianteeman:
Adds help buttons to the components

Do we need to set them up on the help server as well? If so then I dont know how


mbabker:
> Do we need to set them up on the help server as well? If so then I dont know how

Sooner or later pages have to be made on the docs wiki for them all, but the keys all look right so we should be good here.

brianteeman:
Thanks

PS I was going to set them up for the plugins as well but ALL plugins have the same helps screen :(

Sandra97:
@brianteeman It seems that the Help Screens for the 2 Options pages are missing. 
```
JHELP_COMPONENT_USER_ACTIONLOGS_OPTIONS="Components_User_Actionlogs_Options"
JHELP_COMPONENT_PRIVACY_OPTIONS="Components_Privacy_Options"
```

brianteeman:
Thanks - I will fix that shortly

Sandra97:
Another thing: 
`JHELP_COMPONENTS_PRIVACY_REQUEST_EDIT="Components_Privacy_Request_Edit"`
I can't find where it is supposed to be displayed.
When you open a request, it's calling the `JHELP_COMPONENTS_PRIVACY_REQUEST` help screen. So no idea where is the `EDIT` one

mbabker:
That one's when you create a new request.  Mildly confusing by the name alone since the edit layout is actually only ever used to create new requests, but it follows the MVC convention used everywhere else (you're actually on the edit layout when creating new articles too)

brianteeman:
yeah i just followed the existing convention - good or bad

Sandra97:
Thanks.

Sandra97:
Help Screens created. All tested except the 2 'options' one. They work as expected in the backend.
[3.9 Help Screen list](https://docs.joomla.org/Category:Help_screen_3.9)

brianteeman:
The options were merged here #119

PR #101, opened 2018-06-17
Latest Actions module

brianteeman:
Added help string to the module
Corrected english in the strings
Changed the order and width of the columns in the module (and related view in the dashboard) as I think it reads better - see screenshots

### before
chrome_2018-06-17_23-12-50

### after
chrome_2018-06-17_23-20-40


brianteeman:
Thanks

PR #103, opened 2018-06-17
number of days to rotate

brianteeman:
Because it is a number field you could set it to a negative value

I have set a min of 0 and a max of 30 and increments of 5


brianteeman:
thanks

PR #105, opened 2018-06-17
Quickicon messages

brianteeman:
Now that I haave seen this in action I realise that we have to add the word "urgent" otherwise it will confuse people that the count doesnt match the total number of requests

### before
chrome_2018-06-18_00-00-42
chrome_2018-06-18_00-00-46

### after
chrome_2018-06-18_00-01-47
chrome_2018-06-18_00-01-44


brianteeman:
thanks

PR #108, opened 2018-06-17
Dashboard Status Check

brianteeman:
I get what it is intended to do but using a button when there is no action behind it is wrong. It is also confusing because we use that exact same markup for publish/unpublish

### before
chrome_2018-06-18_00-22-09

### after
chrome_2018-06-18_00-44-16


brianteeman:
thanks

PR #109, opened 2018-06-17
Confirm consent

brianteeman:
Unlike the password reset form where this string was copied from this form also has an email field

I am assuming there is logic to that so I have just tweaked the langauge string


brianteeman:
thanks

PR #117, opened 2018-06-19
Fix open privacy article error

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This is an alternative for https://github.com/joomla-projects/privacy-framework/pull/97. When we have privacy consent plugin enabled, users are asked to accept privacy policy before they can access to any pages on the website. The problem with current code is that it doesn't allow users to access to privacy policy article to see the details before accepting. This PR just fixes that logic

### Testing Instructions
1. Make sure plugin System - Privacy Consent is enabled
2. Login to frontend using an account which has not consented privacy policy yet
3. Before patch, you cannot click on Privacy Policy link to see the article
4. After path, you can click on the Privacy Policy link to see the article
5. I tested it myself on multilingual website with associated article, too. But if you can test it, it would be great.

### DIFF
--- a/plugins/system/privacyconsent/field/privacy.php
+++ b/plugins/system/privacyconsent/field/privacy.php
@@ -104,24 +104,9 @@ class JFormFieldprivacy extends JFormFieldRadio
 			$db->setQuery($query);
 			$article = $db->loadObject();
 
-			if (JLanguageAssociations::isEnabled())
-			{
-				$privacyassociated = JLanguageAssociations::getAssociations('com_content', '#__content', 'com_content.item', $privacyarticle);
-			}
-
-			$current_lang = Factory::getLanguage()->getTag();
-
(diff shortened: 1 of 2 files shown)


brianteeman:
Thanks for the pr. I will test it in a few hours when I am on a train to London

IF it works then I am assuming that the same change MIGHT be needed on the new terms plugin

joomdonation:
@brianteeman It only affect if that terms plugin is displayed on edit user profile page in frontend.

brianteeman:
Yes I just realized that. On the other hand it would be good if they use the same code as it makes future maintenance easier. (Assuming there is no negative side effect)

brianteeman:
That works great - thanks

joomdonation:
Thanks @brianteeman. Look like there is a cs error with this PR. Could you please help fixing it? I could not see the error to fix

brianteeman:
I fixed the cs last night with #115

Issue #126, opened 2018-06-20
Non-translated language strings in option=com_actionlogs&view=actionlogs


ReLater:
### Steps to reproduce the issue
- Activate plugin "Action Log - Joomla"
- Do some actions like you can see on this image
![20-06-_2018_17-17-10](https://user-images.githubusercontent.com/20780646/41668040-bf799344-74ae-11e8-95dd-7f48b269cc10.png)
- Check that they are displayed correctly (like image above) in option=com_actionlogs&view=actionlogs

- Deactivate plugin "Action Log - Joomla".
- Check entries in option=com_actionlogs&view=actionlogs again
=> See next image
![20-06-_2018_17-21-40](https://user-images.githubusercontent.com/20780646/41668218-23650104-74af-11e8-8d78-c9f50b2d8028.png)

ReLater:
Similiar issue with module "Latest Actions
![20-06-_2018_17-36-42](https://user-images.githubusercontent.com/20780646/41668877-a9bd21d6-74b0-11e8-863f-73ea3e84d40f.png)
"

joomdonation:
I am not sure if this issue can be fixed. To have correct message displayed, we require the plugin language file loaded. However, in this case, you unpublished the plugin, the language file is not loaded and cause this issue.

ReLater:
Even if it's a small overhead isn't it possible to load the language files of the plugin in the component, too.

I'm not shure if all of these extensions will become locked ones in core. But then all language files should be always present and loadable.

brianteeman:
A logged extension could be uninstalled

ReLater:
> A logged extension could be uninstalled

Yes, but the related actions and related message_language_keys (from plugin actionslog) are still present in table #__action_logs.

I'm just talking about language strings from plg_system_actionlogs here. If it will be a locked extension (I don't know) it should be possible to reliably load these language strings in component or module even if plugin is disabled. As a fallback.

mbabker:
I don't see the issue with loading that plugin's language strings here.  It's a bit of an odd case because components shouldn't depend on a specific plugin, but this is a bit of a special exception (and affect languages versus code behavior, so not strictly the same argument).

brianteeman:
@relater. Sorry I misunderstood which strings it was.

@mbabker should we be using strings from the component and not the plugin then?

mbabker:
That'd probably be better.

joomdonation:
[...]

PR #131, opened 2018-06-21
Display correct action logs even plugins disabled

joomdonation:
Pull Request for Issue #126.
### Summary of Changes
This PR adds a method to load all actionlog plugins language file and calls it each time actions are being displayed to make sure the actions are displayed properly even if the plugins are disabled

### Testing Instructions
1. See https://github.com/joomla-projects/privacy-framework/issues/126 for issue description
2. Apply patch, confirm the issue is sorted.


alikon:
should we consider to be protected ?

administrator/components/com_actionlogs/helpers/actionlogs.php
```diff
@@ -189,4 +189,61 @@ public static function getContentTypeLink($component, $contentType, $id, $urlVar
...
+			)
+			->from('#__extensions')
+			->where('type = ' . $db->quote('plugin'))
+			->where('folder = ' . $db->quote('actionlog'))
+			->where('state IN (0,1)')
+			->order('ordering');
+		$db->setQuery($query);
+		$rows = $db->loadObjectList();
+
+		foreach ($rows as $row)
```
infograf768: what if we have no `$rows` ? Shall not we add a condition?

manojLondhe:
Tested using Joomla patchtester component, this works fine for me.

ReLater:
> should we consider to be protected ?

I would say yes, lock it. Just disabling should be possible.
Ah, I see locking has been pushed to Joomla 3.10 (I hope this feature will be merged then).

PR #133, opened 2018-06-23
[Feature] - Job System, AKA Poor Man's Cron

alikon:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/20839

this system plugin `scheduler` trigger the  `onExecuteScheduledTask` event from the new job plugin group that are ready to be executed depending on their scheduled time

the job plugins must have in their options  3 parameters

- **cachetimeout**  ie how often the task/job plugin event should be triggered
- **lastrun (hidden)** ie the last time the task/job plugin event has been executed
- **unit (hidden)** the unit of measure (mintue =1, hour=3600, days=86400  ect) 

it should be used for (example):
https://github.com/joomla-projects/privacy-framework/blob/88710c2658c9346c84f1ace1389bc0d3f9468cec/plugins/system/privacyconsent/privacyconsent.php

for tasks like: 

- deleteExpiredConsents
- remindExpiringConsents

- logRotation
- updateNotifications

etc...

for  those lucky enough to use a cron job simply schedule
`php cli/scheduler.php`

as a WIP/example  you can look at #217 even if it is a work in progress


Quy:
Missing language file.

brianteeman:
@quy it's an RFC it really doesn't need a code review or language file to
see if it is an appropriate solution

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,267 @@
...
+	protected $autoloadLanguage = true;
+
+	/**
+	 * The scheduler is triggered after the page has fully rendered.
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function onAfterRender()
```
mbabker: Should actually do this `onAfterRespond`.  Plugins doing stuff at `onAfterRender` block sending the HTTP response to the client, so working after the response is sent to the client should be less disruptive to the user who is unlucky enough to have these events/jobs triggered.

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,267 @@
...
+
+		// Abort on failure
+		if (!$result)
+		{
+			return;
+		}
+
+		// check all tasks plugin and if needed trigger those
+		$this->checkAndTrigger();
+
```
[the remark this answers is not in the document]
manojLondhe: Uppercase `// check` => `// Check`

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,267 @@
...
+		{
+			return;
+		}
+
+		if (empty($tasks))
+		{
+			return;
+		}
+
+		// Scatena l'inferno
```
manojLondhe: :+1: :smile:

plugins/system/scheduler/scheduler.php
alikon: done

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,267 @@
...
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	public function onAfterRender()
+	{
+		// Get the timeout for Joomla! system scheduler
+		/** @var \Joomla\Registry\Registry $params */
+		$cache_timeout = (int) $this->params->get('cachetimeout', 1);
+		$cache_timeout = 3600 * $cache_timeout;
```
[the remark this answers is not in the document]
alikon: do we need to add a check for cronnable plugins cachetimeout param that must be > 5 min ?

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,293 @@
...
+		$dispatcher = JEventDispatcher::getInstance();
+
+		foreach ($tasks as $task)
+		{
+			$taskParams = json_decode($task->params, true);
+			
+			$now  = time();
+			$last = (int) $taskParams['lastrun'];
+			$cache_timeout = (int) $taskParams['cachetimeout'];
+			$cache_timeout = 3600 * $cache_timeout;
```
mbabker: Same problem here.  This is again locking the system to only allow these plugins to run too infrequently.  In theory this system needs to be flexible enough to run a task once per minute without forcing plugins that have a shorter timespan (like the update check which defaults to once every six hours) to run too frequently.

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,293 @@
...
+			return;
+		}
+
+		// Check all tasks plugin and if needed trigger those
+		$this->checkAndTrigger();
+		// Log the time it took to run
+		$endTime    = microtime(true);
+		$timeToLoad = sprintf('%0.2f', $endTime - $startTime);
+		JLog::add(
+			'Main Scheduler tooks ' . $timeToLoad . ' seconds',
```
Quy: `took` no s

plugins/system/scheduler/scheduler.php
alikon: solved

plugins/system/scheduler/scheduler.php
alikon: fixed

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,299 @@
...
+		$this->params->set('lastrun', $now);
+
+		$db = JFactory::getDbo();
+		$query = $db->getQuery(true)
+					->update($db->quoteName('#__extensions'))
+					->set($db->quoteName('params') . ' = ' . $db->quote($this->params->toString('JSON')))
+					->where($db->quoteName('type') . ' = ' . $db->quote('plugin'))
+					->where($db->quoteName('folder') . ' = ' . $db->quote('system'))
+					->where($db->quoteName('element') . ' = ' . $db->quote('scheduler'));
+
```
ggppdk: Why 3 indentation Tabs here ? 1 would be enough and allow more visible area for writing code

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,299 @@
...
+		try
+		{
+			// Update the plugin parameters
+			$result = $db->setQuery($query)->execute();
+
+			$this->clearCacheGroups(array('com_plugins'), array(0, 1));
+		}
+		catch (Exception $exc)
+		{
+			// If we failed to execite
```
ggppdk: execute

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,299 @@
...
+		}
+
+		try
+		{
+			// Unlock the tables after writing
+			$db->unlockTables();
+		}
+		catch (Exception $e)
+		{
+			// If we can't lock the tables assume we have somehow failed
```
ggppdk: about comment , should be "unlock" not "lock"

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,299 @@
...
+			->select($db->quoteName(array('extension_id', 'name', 'params')))
+			->from($db->quoteName('#__extensions'))
+			->where($db->quoteName('type') . ' = ' . $db->quote('plugin'))
+			->where($db->quoteName('folder') . ' = ' . $db->quote('cronnable'));
+
+		$db->setQuery($query);
+
+		try
+		{
+			$tasks = $db->loadObjectList();
```
ggppdk: Why not use
`$task = $db->setQuery($query)->loadObjectList();`

administrator/language/en-GB/en-GB.plg_system_scheduler.ini
```diff
@@ -0,0 +1,9 @@
+; Joomla! Project
+; Copyright (C) 2005 - 2018 Open Source Matters. All rights reserved.
+; License GNU General Public License version 2 or later; see LICENSE.txt, see LICENSE.php
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SCHEDULER="System - Job Scheduler"
+PLG_SYSTEM_SCHEDULER_XML_DESCRIPTION="This plugin looks for cronnable plugins to run."
```
brianteeman: Not a fan of cronnable. Is "scheduled" ok to use in this context?

administrator/language/en-GB/en-GB.plg_system_scheduler.ini
```diff
@@ -0,0 +1,9 @@
...
+PLG_SYSTEM_SCHEDULER_CACHETIMEOUT_LABEL="Scheduler (in minutes)"
```
brianteeman: Would this work?

"Frequency (in minutes)"

administrator/language/en-GB/en-GB.plg_system_scheduler.ini
alikon: not sure if i'm correct, but for me, personally speaking, "cronnable" make more clear the meanings

brianteeman: if you use a word that is associated with the "cron" server functionality that isn't actually anything to do with that functionality then it will confuse people

alikon: ok i've no problem with this,
feel free to suggest or direct commit  what is the correct wording in your opinion
a "code owner" surely knows this a lot better than me (an italian with very low level english knowledge)  :smile:

administrator/language/en-GB/en-GB.plg_system_scheduler.sys.ini
```diff
@@ -0,0 +1,7 @@
+; Joomla! Project
+; Copyright (C) 2005 - 2018 Open Source Matters. All rights reserved.
+; License GNU General Public License version 2 or later; see LICENSE.txt, see LICENSE.php
+; Note : All ini files need to be saved as UTF-8
+
+PLG_SYSTEM_SCHEDULER="System - Job Scheduler"
+PLG_SYSTEM_SCHEDULER_XML_DESCRIPTION="This plugin looks for cronnable plugins to run."
```
Quy: change to `scheduler`

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,303 @@
...
+	/**
+	 * Check and Trigger .
+	 *
+	 * @return  void
+	 *
+	 * @since   __DEPLOY_VERSION__
+	 */
+	private function checkAndTrigger()
+	{
+		// Looks for cronnable plugin
```
Quy: change to `scheduler`

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,303 @@
...
+		$db = JFactory::getDbo();
+		$query = $db->getQuery(true)
+			->select($db->quoteName(array('extension_id', 'name', 'params')))
+			->from($db->quoteName('#__extensions'))
+			->where($db->quoteName('type') . ' = ' . $db->quote('plugin'))
+			->where($db->quoteName('folder') . ' = ' . $db->quote('cronnable'));
```
Quy: change to `scheduler`

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,303 @@
...
+			return;
+		}
+
+		if (empty($tasks))
+		{
+			return;
+		}
+
+		// Unleash hell
+		JPluginHelper::importPlugin('cronnable');
```
Quy: change to `scheduler`?

alikon:
`pseudocron` sounds better ?

brianteeman:
For code yes. Not for language strings.

alikon:
maybe **job** is the correct word for both code&languages

plugins/system/scheduler/scheduler.php
```diff
@@ -0,0 +1,303 @@
...
+
+			$startTime = microtime(true);
+			JLog::add(
+				'Main Scheduler:' . $task->extension_id . ':' . $task->name,
+				JLog::INFO,
+				'scheduler'
+			);
+
+			// Update lastrun
+			$this->updateLastRun($task->extension_id, $task->params);
```
mbabker: So, to be honest, these last run checks actually need to be in the plugins being triggered and can't really be efficiently processed here.  It's either that or we don't use the event dispatcher to trigger them.  Not using the dispatcher could have its own advantages as we get into 4.0 with more autoloading capabilities, you could register any callback (well, not a Closure because of serialization quirks) to be the task runner versus the task runners always being plugins (think if you had something that needed to dynamically enqueue a one-time job, I actually do use this type of process on some higher end builds to defer a large task from web to CLI).

As things are right now, you're looping over your result array, processing the last run for each result, and triggering the `onExecuteScheduledTask` event once per row returned.

alikon:
not so sure that i've fully understand....pardon me,  i know it's my bad :smiley: 
but moving to each job plugin the `updateLastRun`  aren't we going to have a lot of repeated code ?

but if we can "dynamically enqueue a one-time job"  this intrigue me  a lot ..... 
even if it is outside of the scope of this silly PR
let me sleep over it a couple  of  fresh nights...oops no cold nights coming soon :smile:

mbabker:
If we use the event dispatcher (`JFactory::getApplication()->triggerEvent('onExecuteScheduledTask');`) to trigger the scheduled jobs, then yes the logic to check when the last run was and mark the last run would have to be duplicated across plugins as they'd need to do something like this:

```php
public function onExecuteScheduledTask()
{
    if (!$this->shouldRun($whateverArgsItTakesToDoThisWell)) {
        return;
    }

    // Do the job

    $this->updateLastRun($whateverParamsAreNeededToWriteAnUpdate);
}
```

If you leave it to the CLI or scheduler plugin to handle that process, you can't use the event dispatcher.  If you import plugins and use the event dispatcher you are ALWAYS going to trigger every listener for that event.  So your PHP code would have to use a different bit of logic...

```php
$jobs = $queryResultFromWhereverItIsYouGetThisDataFrom;

foreach ($jobs as $job) {
    if (!$this->shouldRun($job)) {
        continue;
    }

    if ($job->has_params) {
        // If we get fancy and give a way to store job params...
        $jobParams = $job['params']);

        call_user_func_array($job['callback'], $jobParams);
    } else {
        call_user_func($job['callback']);
    }

    $this->updateLastRun($job);
}
```

jeckodevelopment:
@alikon / @mbabker will this be considered for merge for 3.9?

alikon:
It will be very nice to have in 3.9, but i need to do some changes as
discussed, hope in the weekend I'll come out with a pr on staging, but not
a release blocker

Il ven 21 set 2018, 10:47 Luca Marzo <notifications@github.com> ha scritto:

alikon:
look at https://github.com/joomla/joomla-cms/pull/22310 for the wip

alikon:
new house https://github.com/joomla/joomla-cms/pull/22310

PR #135, opened 2018-06-28
Consent expiration fixes

joomdonation:
Pull Request for Issue # .

### Summary of Changes
This PR fixes some small errors in consent expiration branch (causes SQL errors, 404 not found when click on remind link, token not pre-filled on the form...).

@alikon Please review and merge this change. After this is merged, consider having success human testing from me.


plugins/system/privacyconsent/privacyconsent.php
```diff
@@ -505,21 +508,24 @@ private function remindExpiringConsents()
...
 		$query->where($query->dateAdd($now, $period, 'DAY') . ' > ' . $db->quoteName('created'));
-		$query->where($db->quoteName('remind') . ' = 0');
 
 		try
 		{
-			$users = (int) $db->setQuery($query)->loadObjectList();
+			$users = $db->setQuery($query)->loadObjectList();
 		}
-		catch (JDatabaseException $exception)
+		catch (RuntimeException $exception)
```
mbabker: This one should be reverted so that it's explicitly only catching the exceptions thrown by the database API versus all `RuntimeException`s

joomdonation: According to this doc block https://github.com/joomla-projects/privacy-framework/blob/dev/privacy/libraries/joomla/database/driver.php#L1682 , it throws RuntimeException and that's why I changed it. So the doc block is wrong?

mbabker: It's right but not 100% accurate.  The specific Exception subclasses all extend RuntimeException (otherwise we would've broken B/C by changing the type of Exception) and JDatabaseException is specifically an interface for all of those subclasses.  So it's right in that a RuntimeException is thrown, but it's wrong in that it's not specifically RuntimeException anymore but depending on the error case one of three subclasses which can be caught with the JDatabaseException interface.

joomdonation: OK. I changed it back to JDatabaseException

Issue #136, opened 2018-06-28
Action log for non-confirmed user


webgras:
### Steps to reproduce the issue
Setup:
In User Global Options:
Set "New User Account Activation" to "Self"

In frontend:
Register yourself as a new FE user
(email with confirmation link will be sent -->don't confirm!)
(User is not enabled and not activated)
Try to login in frontend
Frontend warning: Warning - Login denied! Your account has either been blocked or you have not activated it yet.

### Expected result
User Action Log in Backend:
"User user2 (not activated) tried to log in to site"

### Actual result
User Action Log in Backend:
"User user2 logged in to site"

### System information (as much as possible)
Joomla! 3.9.0-dev Development [ Amani ] 26-June-2018 15:45 GMT

### Additional comments
As for spam protection, I also tried a non-existing user --> no action log --> as I expected! :)

mbabker:
The action log plugin should actually hook the `onUserAfterLogin` and `onUserLoginFailure` events instead of only `onUserLogin`.

`onUserLogin` is when an account being blocked is set by the user plugin, `onUserAfterLogin` is fired for a successful login, and `onUserLoginFailure` fired for a failed login.

// @alikon @joomdonation

joomdonation:
PR https://github.com/joomla-projects/privacy-framework/pull/142 should address this issue

alikon:
closing as we have a pr #142

PR #142, opened 2018-06-29
Improve login action log

joomdonation:
Pull Request for Issue #136 .

### Summary of Changes
This PR makes some improvements to log action login feature:

- Use onUserAfterLogin instead of  onUserLogin event trigger to log success login action
- Implement onUserLoginFailure to log failed login action

### Testing Instructions
- Apply path
- Try to login using a valid username and password, check action log and make sure the login action is logged
- Try to login using a valid username and wrong password, check action log and make sure there is action is logged (Example User admin tried to login to site)


plugins/actionlog/joomla/joomla.php
```diff
@@ -757,9 +756,47 @@ public function onUserLogin($user, $options = array())
...
+
+		if (!$this->checkLoggable($context))
+		{
+			return;
+		}
+
+		$loggedInUser       = JUser::getInstance($response['username']);
+
+		// Not a valid user, return
+		if ($loggedInUser->id)
```
alikon: should be
`if (!$loggedInUser->id)`

joomdonation: Yes, changed. Thanks

PR #143, opened 2018-06-29
Escaping doublequotes in en-GB.plg_system_actionlogs.ini

infograf768:
As title says.
Can be merged on review


joomdonation:
Can we use single quote like how we use it en-GB.plg_actionlog_joomla.ini https://github.com/joomla-projects/privacy-framework/blob/dev/privacy/administrator/language/en-GB/en-GB.plg_actionlog_joomla.ini#L35 ?

infograf768:
We could, but why not using doublequotes as we can.
I know that in French language I prefer to use the single quotes elsewhere than in href in order to separate them clearly.

PR #146, opened 2018-06-29
Fix save User Action Log Options error

joomdonation:
Pull Request for Issue # .

### Summary of Changes
As of right now, when we edit an account, the User Action Log Options could not be saved. This PR fixes that error.

### Testing Instructions
1. Login using a super admin account
2. Try to edit an account and change Action Log Options
3. Before patch, the change could not be saved
4. After path, the change will be saved.


joomdonation:
Question: Since  we only want to allow this setting for Super Users, maybe we should only show this setting on edit user form, not when create new user?

joomdonation:
Thanks @SharkyKZ. Not sure why we don't pass $data to onContentPrepareForm on save. I use the suggested workaround for now as it's the safest option.

alikon:
for me too

PR #156, opened 2018-07-05
Various fixes to com_actionlogs

Hackwar:
### IP filtering
Right now the IP address is filtered and all non-numeric characters are removed. This sets the Input filter to raw and then applies filter_var() on the result.

### CSV export
Some messages weren't translated when doing the CSV export because the translation files are loaded when translating the extension name, but that happened only after the messages were translated.

### Improve list layout
The list layout still had some older code. This should get this up to par with the rest of the core components. Meh, maybe not 100%...

Just wanted to provide at least one small first PR. :smile:


administrator/language/en-GB/en-GB.com_actionlogs.ini
```diff
@@ -23,6 +23,7 @@ COM_ACTIONLOGS_EXTENSION_FILTER_DESC="Search in the User Action logs by extensio
 COM_ACTIONLOGS_IP_ADDRESS="IP Address"
 COM_ACTIONLOGS_IP_ADDRESS_ASC="IP Address ascending"
 COM_ACTIONLOGS_IP_ADDRESS_DESC="IP Address descending"
+COM_ACTIONLOGS_IP_INVALID="IP invalid"
```
brianteeman: Typical English and not following any logic. Please can you change to 

Invalid IP

Hackwar:
I renamed addLogsToDb to addLog now, too.

administrator/language/en-GB/en-GB.com_actionlogs.ini
Hackwar: done

Issue #159, opened 2018-07-06
Wrong link in Latest Actions panel


Hackwar:
From comment by @neo191987 https://github.com/joomla/joomla-cms/pull/20800#issuecomment-403025252

Hi, I'm a new tester here. I found several problems in the new feature "Latest Actions"
When I create a menu item from the administration in the module "Latest Actions", will show the following:
User test trashed the menu item Create Request, Confirm Request or Extend Consent.
Clicking on the links: Create Request, Confirm Request or Extend Consent
I will get this error: Invalid controller: name='menu_item', format=''

Hackwar:
Closing since we have a (working) PR #183

PR #183, opened 2018-07-07
Fix wrong items link in action logs

joomdonation:
Pull Request for Issue #159.

### Summary of Changes
This PR fixes wrong links to item edit page of several content types in action logs (Menu item, access level, banner client). Basically, instead of using type_title (menu_item, access_level, banner_client), we use type_alias (com_menus.item, com_users.level, com_banners.client) to detect content type to generate link to item edit page of each content type.

### Testing Instructions
1. Add new access level
2. Access to action logs component, note that the link to access level is causing error
3. Apply patch, add new access level, confirm that the link to that access level in action logs is now correct
4. Add new article, then check and make sure link to that article in action log is correct, too.


Hackwar:
Tested and works as expected. Do we need 2 testers?

Closing #159, since we have this working PR.

alikon:
i think your test is more than enough :smile:

PR #185, opened 2018-07-07
Make sure the language of com_privacy is loaded as it is not running as plugin

zero-24:
Pull Request for Issue https://github.com/joomla/joomla-cms/pull/20800

### Summary of Changes

Make sure the language of com_privacy is loaded as it is not running as plugin. I'm not sure if that is the right spot. But even in the helper class it is not correct or what do you think?

### Testing Instructions

Do some com_privacy actions

### Expected result

image

### Actual result

image

None


mbabker:
I'd almost suggest `ActionlogsHelper::loadActionLogPluginsLanguage()` should have this call in it.

zero-24:
implemented as suggested.

Issue #192, opened 2018-07-08
Privacy Policy Message in Register Form


coolcat-creations:
<img width="688" alt="grafik" src="https://user-images.githubusercontent.com/828371/42422910-20dd603a-82f0-11e8-83c7-2bbf561f5cd0.png">

Would be great to be able to use JTEXT here for other languages.

brianteeman:
would be great but not possible. It's an issue with joomla and how the field was coded originally see #12793

coolcat-creations:
but how can one insert here text in the sitelanguage? If you refer to https://github.com/joomla/joomla-cms/issues/12793 thats not the default value i am speaking about. I put in a PRIVACY_MY_TEXT string and added it into the overrides but it's not translated.

coolcat-creations:
I see, so it IS translateable but was coded that way that I can't use an own string. I think It should be easy to use an own string, because it's not about #12793 - you can just wrap a JTEXT around the output and it will be translated if an override is available.

brianteeman:
as explained on your pr that doesnt work

coolcat-creations:
How did you test this for me it did work ?

brianteeman:
yes sorry I was getting confused with #12793 That "bug" was the reason that its a hint and not a default value.

PR #193, opened 2018-07-08
Update privacy.php

coolcat-creations:
Add possibility for own JTEXT Key

Pull Request for Issue #192 

### Summary of Changes
Wrapped the Message in JTEXT so the user can define an own string too.

### Testing Instructions
Edit the Privacy Consent Plugin and insert a MY_CONSENT_TEXT string into the message box
Add overrides for your languages

### Expected result
The Registration form shows your translated strings even if you don't use the official string

### DIFF
--- a/plugins/system/privacyconsent/field/privacy.php
+++ b/plugins/system/privacyconsent/field/privacy.php
@@ -40,7 +40,7 @@ class JFormFieldprivacy extends JFormFieldRadio
 	{
 		$privacynote = !empty($this->element['note']) ? $this->element['note'] : Text::_('PLG_SYSTEM_PRIVACYCONSENT_NOTE_FIELD_DEFAULT');
 
-		echo '<div class="alert alert-info">' . $privacynote . '</div>';
+		echo '<div class="alert alert-info">' . Text::_($privacynote) . '</div>';
 
 		return parent::getInput();
 	}


brianteeman:
but this prevents you simply typing in some content forcing everyone to create a language string override even on a single language site

coolcat-creations:
No, you can type in a simple text too- No need to create a language string

brianteeman:
um - yes i see - that but how is this different to the current instructions

> Note for Multilingual sites. If you use the default text then it will be displayed in the user's language. It is not possible to translate the custom text. If you wish to customise the text and display it in multiple languages then you should leave this field blank and use the Joomla language override facility to customise the PLG_USER_PRIVACYCONSENT_NOTE_FIELD_DEFAULT language string for each language installed.

brianteeman:
I must be missing something here - because to me you are just offering a way to create a new constant that needs to be overridden instead of simply using the already defined constant

coolcat-creations:
It's adding few characters of code to make it more flexible. The defined constand can be always the fallback but that way was my intuitive attempt to use that field...

brianteeman:
sorry I still dont get what is different

mbabker:
To me it's extra flexibility with needless complexity.  To me it's a more consistent experience if you're translating the same constant on every site versus having to lookup the reference constant to translate on a per-site basis.  So I honestly don't think there's much to be gained here.

brianteeman:
This functionality would have to be added to every field that accepts input other wise people will think they can enter language constants there. Consistency has to win. So sorry but I am closing this

coolcat-creations:
Ok, but then you should mention the right constant on that page instead of having the user look it up in the documentation.

brianteeman:
No, for so many reasons, that we already discussed everywhere in joomla. The documentation is the place for this not the UI

PR #53, opened 2018-05-27
Export Data from Request List

mbabker:
Pull Request for Issue #3

### Summary of Changes

This adds the architecture to collect data for an export information request and give the site owner a downloadable document.  Notable changes:

- Extends `Joomla\CMS\Document\XmlDocument` to configure whether it serves an inline document or an attachment (downloads the file)
- Adds the `onPrivacyExportRequest` event which receives one parameter, the `PrivacyTableRequest` object representing the request being processed
- Adds the `plg_privacy_user` plugin to process the core user data

### Testing Instructions

With this branch/patch installed and a valid export request in place, hit the export button on the requests list view to process the request and download the data, you should get a valid XML document with all available data.

#### Plugin Event Handler

It is the responsibility of plugins listening to the `onPrivacyExportRequest` event to create and return an array of `PrivacyExportDomain` objects containing all of the data to be exported.  As seen in the `PrivacyExportDomain` class, it and the object it uses are all rather simple data transfer objects to enforce a consistent data structure.  Similar to the capability reporting system, these plugins are read and append only, so they are not able to modify the information returned by other plugins.

#### Privacy Plugins

It is suggested that privacy plugins extend the `PrivacyPlugin` class in the component.  This gives access to a bunch of shortcut helpers to create the data objects used for the export process.  This isn't a hard requirement but if you want to do these functions yourself then more power to you.

#### XML Document

The XML document that is generated (and some of the terminology) is very loosely modeled on the Akeeba data compliance component structure primarily because the use of XML here is going to be one of the most flexible options since items can have an indeterminate number of fields so it makes use of "standardized" formats (CSV for example) impractical.

#### Model Data Return

Purposefully I'm having the MVC model class return the collection of data objects versus the model (and inherently the plugins) being forced into a single format.  This leaves a little flexibility in how the exported data is processed (i.e. someone could come along and write a plugin generating a HTML document that's a bit more user friendly to consume).

PR #196, opened 2018-07-09
a11y requests

brianteeman:
Make sure the links to export data, email data etc are announced correctly to a screenreader

They probably should be buttons as well and not a links styled as buttons but for some reason I couldnt get that to work


mbabker:
> They probably should be buttons as well and not a links styled as buttons but for some reason I couldnt get that to work

I tried once upon a time.  Let's just say the core UI tooling doesn't handle some of the cases we're throwing at it here so instead of trying to explain why one "button" is a link and the rest are "proper" buttons everything is just button styled links (https://github.com/joomla-projects/privacy-framework/pull/53#issuecomment-392591464 and https://github.com/joomla-projects/privacy-framework/pull/53/commits/60b8c2969fa02aa23760dfcc727b1a889058a2f8)

brianteeman:
Thanks

Issue #172, opened 2018-07-06
com_actionlogs: improve export filename


Hackwar:
From this comment by me https://github.com/joomla/joomla-cms/pull/20800#issuecomment-402661866

The filename of the CSV export has UTC time, which I can understand why, but then we should add UTC in front of it.

brianteeman:
Feel free to submit a pr to do that - can't see the need myself but if you want to then go for it

joomdonation:
I don't see the need myself, too. However, If I were modified it, I would use current user timezone because it's easier for him to see when the it was exported by looking at filename

Hackwar:
I thought the user timezone might be good, too, but then I thought about inter-timezone disputes. If we have an admin and a disputee with differing timezones, then storing the file with the timezone of one of the two might create more confusion, so I would just try to make clear that the timestamp in the filename is in UTC.

Hackwar:
I added "UTC" to the filename in #197. What do you think about this? If I get more than one downvote, I'll simply close PR and issue. Don't need to waste more time on this than absolutely necessary. :wink:

alikon:
not a downvote from my but imo not so a priority ...

Hackwar:
To cleanup, I'm closing this issue. The PR is where we should discuss this further and decide.

PR #197, opened 2018-07-09
Adding UTC timezone to filename for export

Hackwar:
PR for #172 
This is an attempt to make the timezone of the timestamp in the filename of the export of CSV logs apparent.

### DIFF
--- a/administrator/components/com_actionlogs/controllers/actionlogs.php
+++ b/administrator/components/com_actionlogs/controllers/actionlogs.php
@@ -73,7 +73,7 @@ class ActionlogsControllerActionlogs extends JControllerAdmin
 		if (count($data))
 		{
 			$rows = ActionlogsHelper::getCsvData($data);
-			$filename     = "logs_" . JFactory::getDate();
+			$filename     = "logs_" . JFactory::getDate() . 'UTC';
 			$csvDelimiter = ComponentHelper::getComponent('com_actionlogs')->getParams()->get('csv_delimiter', ',');
 
 			$app = JFactory::getApplication();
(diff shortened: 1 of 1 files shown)


brianteeman:
If you think it matter then fine by me

administrator/components/com_actionlogs/controllers/actionlogs.php
```diff
@@ -73,7 +73,7 @@ public function exportLogs()
 		if (count($data))
 		{
 			$rows = ActionlogsHelper::getCsvData($data);
-			$filename     = "logs_" . JFactory::getDate();
+			$filename     = "logs_" . JFactory::getDate() . 'UTC';
```
brianteeman: Mr non-programmer speaking but shouldnt it be

>			$filename     = "logs_" . JFactory::getDate() . "UTC";

Hackwar: Civil war incoming... I prefer the single quotes and we are using it in most cases in Joomla, so I used those. This depends quite a bit on your personal preference...

brianteeman: It was just that the prefix has double and the suffix single and consistency always wins for me. I have no idea what our style guide says

mbabker: Personally I'd rather this be explicitly formatted via `JFactory::getDate()->format()` instead of relying on `__toString()`, especially as the default `Y-m-d H:i:s` isn't the most filesystem friendly name for some systems.

Hackwar:
Changed the formatting to be more FS friendly.

PR #198, opened 2018-07-09
Adding missing # in table name

Hackwar:
The sample data is missing a #. I think this is pretty self-explanatory...


brianteeman:
Good spot

alikon:
please do the same fix on postgresql etc

Hackwar:
done

Issue #173, opened 2018-07-06
com_actionlogs: Edits to global configuration are not logged


Hackwar:
From this comment by me https://github.com/joomla/joomla-cms/pull/20800#issuecomment-402661866

Edits to Global Configuration are not logged

joomdonation:
If I understand your issue correctly, it should be sorted by PR https://github.com/joomla-projects/privacy-framework/pull/146

Hackwar:
I don't think so. I'm talking about the site configuration in com_config. In my tests, changes to a components configuration was correctly saved in the log, but changes to configuration.php through the GUI was not.

joomdonation:
Ah, I mis-understood the issue

joomdonation:
So I tried and it's working OK to me. 

1. Access to System -> Global Configuration

2. Select User Actions 

3. Tried to change the parameters, it's being saved properly

So I'm not sure what's wrong here?

mbabker:
The actual global configuration.  Not "go to global configuration then choose a component to change a component's configuration".

Hackwar:
We have a PR #200

PR #200, opened 2018-07-10
Adding logging for global configuration

Hackwar:
When editing the global configuration (of the application, not a specific component), this is not logged in the actionlog. This PR implements that.


Hackwar:
This solves #173

administrator/components/com_config/model/application.php
```diff
@@ -395,7 +396,12 @@ public function save($data)
 		$this->cleanCache('_system', 1);
 
 		// Write the configuration file.
-		return $this->writeConfigFile($config);
+		$result = $this->writeConfigFile($config);
+
+		// Trigger the after save event.
+		$dispatcher->trigger('onApplicationAfterSave', array($config));
```
joomdonation: Should we change event name to **onApplicationConfigAfterSave** ? I think it's more clear than  onApplicationAfterSave

plugins/actionlog/joomla/joomla.php
```diff
@@ -288,6 +288,53 @@ public function onContentChangeState($context, $pks, $value)
...
+
+		// Not found a valid content type, don't process further
+		if ($params === null)
+		{
+			return;
+		}
+
+		$extensionType = $params->type_title;
+
+		$messageLanguageKey = strtoupper('PLG_ACTIONLOG_JOOMLA_' . $extensionType . '_UPDATED');
```
joomdonation: Personal, I would use $messageLanguageKey = 'PLG_ACTIONLOG_JOOMLA_APPLICATION_CONFIG__UPDATED'; directly instead of getting it from $params. On other methods in this plugin (like onContentAfterSave), we have to get $extensionType via $params variable because there are multiple extensions/ content types use that same event. For this event, only com_config use it, so I think $extensionType could be hardcoded to APPLICATION_CONFIG

administrator/components/com_config/model/application.php
Hackwar: The event for extensions is named ```onExtensionAfterSave```, so I rather would want to stick to that format.

mbabker:
Since we're adding the after event, should we add the corresponding before event as well?  Or does it not make sense to have a pre-save hook in this context?

mbabker:
Also, the `onXXX(Before|After)Save` events usually take a set of three arguments:

- context
- `JTable` object with data
- `$isNew` flag

Do we want to try and follow convention or say "screw it" and stick with this event signature?

Hackwar:
Since there is no context, no JTable and no new config, I'd say screw it.

Regarding the before-event: Yes, we should add this. I was just lazy... I'm going to add that soon.

Hackwar:
I added the events and updated the table names for the update SQLs.

Issue #222, opened 2018-07-30
Missing value in  Plugins: Privacy - Content


Sandra97:
The value is missing for this constant: 
![index](https://user-images.githubusercontent.com/4682328/43395597-eb77a55c-93fe-11e8-962a-63aebe86dc68.png)

alikon:
please test #223

PR #223, opened 2018-07-30
Issue #222 fix: Missing value in Plugins: Privacy - Content

manojLondhe:
Fixed typo


Sandra97:
I just applied it. It works as expected, the value of the constant is displayed.

PR #225, opened 2018-08-07
Refactoring logic to create privacy requests

Hackwar:
Pull Request for Issue #158.

### Summary of Changes
When creating a request, you were able to give a user and a mail address that were not connected/related/whatever. This PR forces that either no user is assigned to a request or the user has to have the same email address as the given email address. Or no email address in the frontend is given, then the logged in user is used.

Please check the language strings. I'm not really happy about them, but maybe someone can come up with something better? @brianteeman


administrator/language/en-GB/en-GB.com_privacy.ini
```diff
@@ -63,13 +63,16 @@ COM_PRIVACY_ERROR_CANNOT_REMOVE_DATA="The data for this request cannot be remove
 COM_PRIVACY_ERROR_CANNOT_REMOVE_UNCONFIRMED_REQUEST="The data for an unconfirmed request cannot be removed."
 COM_PRIVACY_ERROR_COMPLETE_TRANSITION_NOT_PERMITTED="This record cannot be updated to \"Completed\" status."
 COM_PRIVACY_ERROR_EXPORT_EMAIL_FAILED="Export failed with the following error: %s"
+COM_PRIVACY_ERROR_GIVEN_USER_IS_INVALID="The given user does not exist!"
```
brianteeman: provided not given

administrator/language/en-GB/en-GB.com_privacy.ini
```diff
@@ -63,13 +63,16 @@ COM_PRIVACY_ERROR_CANNOT_REMOVE_DATA="The data for this request cannot be remove
...
 COM_PRIVACY_ERROR_INVALID_TRANSITION_NOT_PERMITTED="This record cannot be updated to \"Invalid\" status."
+COM_PRIVACY_ERROR_NO_USER_OR_EMAIL_GIVEN="You have to provide either an email-address or a user."
 COM_PRIVACY_ERROR_REMOVE_DATA_FAILED="Removal failed with the following error: %s"
 COM_PRIVACY_ERROR_REQUEST_ID_REQUIRED_FOR_EXPORT="A request ID is required to export data."
 COM_PRIVACY_ERROR_REQUEST_ID_REQUIRED_FOR_REMOVE="A request ID is required to remove data."
 COM_PRIVACY_ERROR_REQUEST_TYPE_NOT_EXPORT="Only export requests may be exported."
 COM_PRIVACY_ERROR_REQUEST_TYPE_NOT_REMOVE="Only removal requests may have their data removed."
 COM_PRIVACY_ERROR_UNKNOWN_REQUEST_TYPE="Unknown information request type."
+COM_PRIVACY_ERROR_USER_EMAIL DOES_NOT_EQUAL_GIVEN_EMAIL="The email-address does not belong to the account of the given user."
```
brianteeman: provided not given

administrator/language/en-GB/en-GB.com_privacy.ini
```diff
@@ -63,13 +63,16 @@ COM_PRIVACY_ERROR_CANNOT_REMOVE_DATA="The data for this request cannot be remove
 COM_PRIVACY_ERROR_CANNOT_REMOVE_UNCONFIRMED_REQUEST="The data for an unconfirmed request cannot be removed."
 COM_PRIVACY_ERROR_COMPLETE_TRANSITION_NOT_PERMITTED="This record cannot be updated to \"Completed\" status."
 COM_PRIVACY_ERROR_EXPORT_EMAIL_FAILED="Export failed with the following error: %s"
+COM_PRIVACY_ERROR_GIVEN_USER_IS_INVALID="The given user does not exist!"
 COM_PRIVACY_ERROR_INVALID_TRANSITION_NOT_PERMITTED="This record cannot be updated to \"Invalid\" status."
+COM_PRIVACY_ERROR_NO_USER_OR_EMAIL_GIVEN="You have to provide either an email-address or a user."
```
brianteeman: email address

administrator/language/en-GB/en-GB.com_privacy.ini
brianteeman: email address

brianteeman: missing _ in the constant

TobsBobs: and email address :)

Hackwar:
Thank you @brianteeman!

administrator/components/com_privacy/models/request.php
```diff
@@ -389,6 +389,37 @@ public function validate($form, $data, $group = null)
...
+			if (!$user->id)
+			{
+				$this->setError(JText::_('COM_PRIVACY_ERROR_GIVEN_USER_IS_INVALID'));
+
+				return false;
+			}
+
+			if ($validatedData['email'] && $user->email != $validatedData['email'])
+			{
+				$this->setError(JText::_('COM_PRIVACY_ERROR_USER_EMAIL DOES_NOT_EQUAL_GIVEN_EMAIL'));
```
mbabker: Missing `_` in the constant here.

administrator/components/com_privacy/models/request.php
```diff
@@ -389,6 +389,37 @@ public function validate($form, $data, $group = null)
...
+			$user = JFactory::getUser($validatedData['user_id']);
+
+			if (!$user->id)
+			{
+				$this->setError(JText::_('COM_PRIVACY_ERROR_GIVEN_USER_IS_INVALID'));
+
+				return false;
+			}
+
+			if ($validatedData['email'] && $user->email != $validatedData['email'])
```
mbabker: Pedantic nitpicking here but most of the stuff I did in this component uses strict comparisons (`!==` or `===`), please follow suit.

components/com_privacy/models/request.php
```diff
@@ -62,9 +62,26 @@ public function createRequest($data)
...
 		$user = JFactory::getUser();
 
-		if (!$user->guest)
+		if ($user->guest && !$data['email'])
+		{
+			$this->setError(JText::_('COM_PRIVACY_ERROR_NO_MAIL_OR_USER_GIVEN'));
+
+			return false;
+		}
+		else
```
mbabker: Since you're returning in the above if you don't need the else here.

Hackwar:
I did the required changes. Hopefully everything is fine now. :smile:

alikon:
suppose i log-in 
create an export request, giving a different e-mail address
confirm the export request
now the request is in a confirmed status but  with Associated User:N/A 

now from backend  work that request (click on export data)
you got something like
```xml
<?xml version="1.0"?>
<data-export>
  <domain name="user contact" description="Joomla! user contact data"/>
</data-export>
```
that seems pointless to me

shouldn't the request be associated (always) to the logged-in  user  instead of N/A ?

Hackwar:
Why should I be able to file a request where the user and the provided mail address are different from each other? I could come up with a dozen ways to exploit that then.

Either provide the correct mail address or leave it empty, to associate the request with the current user.

brianteeman:
Always go for the simple option

Hackwar:
I didn't even know that the form is not available for guests... I guess I wont be the only one with that issue...

I don't know which way to go, but we can't leave it like it is right now. I hope we can agree on that. This PR currently has the most permissive implementation of all possibilities that we discussed here, so my personal preference would be to go with this PR and refine this in 3.9.1, mostly because GDPR took effect in May and we now have August. If that is not acceptable, then please tell me which solution I should implement. Unless I'm mistaken, we have the following possibilities:

* Only logged-in users can create requests and thus we don't need to collect the mail-address at all. Simply click a button to create a request in the frontend and in the backend only select a user. The whole mail-address thingy is removed/commented out.
* Everybody can create a request, both for users and for mail-addresses as identifier. Logged-in users only get to see a button, anonymous users get a form for a mail-address. (Mostly this PR)
* Mail-addresses are checked against our logged in users and the association to a user is created automatically.

What should I do?

mbabker:
So to be clear, the current implementation isn't open to exploitation in any way.  The current request form (only accessible after login) basically says "I am user X and request all data related to my account to be exported or removed and I would like notifications of this action (or the data export if that is the request type) to go to this email address which I will use to also confirm access to the account and the validity of this request".  It does not open the door to Hannes logging onto a site, putting my email address in there, and getting my data.  If that is how anyone has interpreted this then it is dead wrong.

To be quite honest, if we're going to change anything we should just drop the relations from privacy requests to user accounts.  Everyone puts in an email, and whatever is directly related to that email is what is returned.  Then there's no way anyone can spin this around to say "oh but if I'm logged in my account and I put my BFF's email in there then there's a security loophole in this system and OMGWTFBBQ you all suck!".

mbabker:
Superseded by https://github.com/joomla-projects/privacy-framework/pull/233

PhilETaylor:
How to exploit this to cause issues:

1: create new joomla user,
2: login
3: get token
4: post unlimited repeatedly a request to com_privacy with different email addresses
5) crash server with DDoS/HardDisk full.

There is no flood protection.

Issue #177, opened 2018-07-06
Codereview com_actionlogs helper


Hackwar:
From this comment by me https://github.com/joomla/joomla-cms/pull/20800#discussion_r199324065

The getDataByPks() method is such a general method, that it feels wrong here.

Hackwar:
Also from this comment https://github.com/joomla/joomla-cms/pull/20800#discussion_r199324036
And https://github.com/joomla/joomla-cms/pull/20800#discussion_r199323965
The comments are strange

alikon:
if i'm understood well  it's only a naming convention issue so what it should be  ?
`getDataByActions` ?
consider we are only calling in this way `ActionlogsHelper::getDataByPks` 
https://github.com/joomla-projects/privacy-framework/blob/dev/privacy/plugins/actionlog/joomla/joomla.php#L275

joomdonation:
Actually, the method is only used to get title of published/unpublished items (categories, articles...) and add that title to each action log record (getting $pk from the method can be removed as it's not used at the moment)

How about rename it to **getItemsTitle** ?

Hackwar:
That method is retrieving an arbitrary field with the corresponding ID from an arbitrary table. You could name this Database::loadValueFromTable() and it would be fine, but here this is such a general method, that I think it is wrong here. What is keeping you from retrieving the title of an article with this? Or the parameters of a contact? This is simply not com_actionlogs specific code, but something that I would (in worst case) expect as part of an ORM or database library.

Hackwar:
Closing in favour of #226.

PR #226, opened 2018-08-12
Removing getDatabyPks()

Hackwar:
Pull Request for Issue #177.

This removes ActionlogsHelper::getDataByPks(). That method is used exactly once and it is such a generalised method, that it simply does not belong into a component helper like this. You could get the passwords of users with this method as well as the message bodies of com_messages. Either add such a method to the database layer or create a method that is somehow specific to com_actionlogs, but the way it is now, we should simply put the code into the joomla plugin, as I have done here.


Hackwar:
Added the try-catch back in.

Issue #228, opened 2018-08-15
Duplicate Privacy menu items in left sidebar


Sulpher:
We can see two new menu items in Menu:
 - Privacy
 - User Actions log

But there are no such menu items in the left sidebar.
My proposal is to duplicate them there. 

![screen shot 2018-08-15 at 13 53 09](https://user-images.githubusercontent.com/3432048/44145174-3a163944-a093-11e8-8f93-d9f4a2047b83.png)

Sulpher:
@alikon I have been tested these changes: all works fine and the task is resolved. Thank you!
![screen shot 2018-08-16 at 21 38 45](https://user-images.githubusercontent.com/3432048/44228002-c6f95800-a19c-11e8-9c56-991905002f18.png)

PR #229, opened 2018-08-16
add privacy & actionlogs to sidebar

alikon:
Pull Request for Issue #228 .

### Summary of Changes
add privacy & actionlogs to the sidebar


mbabker:
To be honest, I'm not sure I actually agree with this one.  To me, it can be argued that the full com_users sidebar should be shown on both com_actionlogs and com_privacy, and the full com_privacy sidebar be shown on both com_actionlogs and com_users.  I don't think these two components need to be linked from the com_users sidebar, we already don't do this for the mass mail page.

alikon:
i've no problem with or without this ....just a silly pr to answer to an open issue 
feel free to close it

PR #230, opened 2018-08-16
[wip] check email sended before to store the request

alikon:
Pull Request for Issue https://github.com/joomla/joomla-cms/issues/21646 .

### Summary of Changes
just send/check the email before to store the request

### Testing Instructions
https://github.com/joomla/joomla-cms/issues/21646

###  Actual result
request stored but email not sended

###  Expected result
At least request not stored if email not sended/enabled


Hackwar:
This will not work, since we need the ID of the created row in the code to send the mails. I disagree that we have to first send the mail and then create the request. First make sure that the data is stored securely in our DB, then send out mails.

alikon:
maybe i'm wrong but where we use the request ID in the email ?

mbabker:
The ID's not used but if the record fails to save then the email's useless because there's nothing to validate the token from the email against.

alikon:
so we are in the "_Chicken or the egg_" dilemma

Hackwar:
Sorry, I misread the code. We are indeed not using the ID. But as Michael said, the issues are greater that a write fails than that the mail is not sent. If the mail is sent and the write fails, people would assume that it worked out because they got a confirmation.

mbabker:
Not really.

Everything I pointed out in https://github.com/joomla/joomla-cms/issues/21646#issuecomment-413655824 is really what this PR should be.

Mail sending should still be after the record is saved in the database (basically the changes to the model reverted as things stand right now).  It is still entirely possible to save a valid request and in the event the mail send fails something could be wired up to try sending the message for that request again instead of flat out abandoning it.  Now, you have the situation of if the mail sending succeeds but the database update fails there is no salvaging things at all.

I will work on the code solution for this tomorrow.

alikon:
>  if the mail sending succeeds but the database update fails there is no salvaging things at all.

database never fails 😄

alikon:
i'm closing this in favour of your upcoming pr
i've added the dashboard check in #231

PR #233, opened 2018-08-19
Remove association to user accounts in privacy requests

mbabker:
### Summary of Changes

The association of privacy requests to user accounts has created a rather controversial response with a fair bit of hypothetical "but I can use a non-associated email address to request data for my account" scenarios.  To stop that discussion, linking requests to user accounts is now just flat out removed.

Going forward, whatever email address you put in (when authenticated or not, if not authenticated is ever supported on the frontend) is what email address will be processed (after it is validated).  If an account exists for that email, it will be loaded in and acted upon as appropriate.

### Testing Instructions

Requests can continue to be processed.

Doubtful, I don't think this part has been written yet.


administrator/components/com_privacy/models/remove.php
```diff
@@ -63,12 +63,24 @@ public function removeDataForRequest($id = null)
 			return false;
 		}
 
+		// If there is a user account associated with the email address, load it here for use in the plugins
+		$userId = (int) $this->getDbo()->setQuery(
+			$this->getDbo()->getQuery(true)
+				->select('id')
+				->from($db->quoteName('#__users'))
```
alikon: same here

administrator/components/com_privacy/models/request.php
```diff
@@ -254,9 +254,18 @@ public function notifyUserAdminCreatedRequest($id)
 
 		$lang = JFactory::getLanguage();
 
-		if ($table->user_id)
+		$userId = (int) $this->getDbo()->setQuery(
+			$this->getDbo()->getQuery(true)
+				->select('id')
+				->from($db->quoteName('#__users'))
```
alikon: here

mbabker:
Should be all fixed up now.

alikon:
ok errors fixed

but still a bit inconsistent imho
if you suppose to have a confirmed remove request with a non-associated email address

screenshot from 2018-08-21 19-20-27

no data has been removed cause of non-associated email address

mbabker:
> if you suppose to have a confirmed remove request with a non-associated email address
> 
> no data has been removed cause of non-associated email address

Well, the only two things in core I can think of to actually sanitize/remove in core using just an email are either the user account or a contact record.  So, this is basically the expected behavior until you get some other plugins in place that might do something meaningful with anonymous user data.

alikon:
For me we should choose from  2 behaviuor:
- let things flow as before this PR 
or 
- don't allow non-associated email address request to be submitted/processed

mbabker:
> let things flow as before this PR

It does.  The only practical difference is the pseudo foreign key in the requests table to the users table is no longer present and to create an association between users and requests you have to do a lookup based on email address.

> don't allow non-associated email address request to be submitted/processed

Not an option.  Same example I always use, commenting systems where you don't have to be logged in to submit a comment (but you still submit your name and email address).  If we restrict this system to only support user accounts on the website we massively restrict the way this system can be used and would never be able to support use cases like the person commenting without creating an account on the website.

alikon:
ok let's move on then

PR #235, opened 2018-08-24
Make sure we don't send a referrer on the request

alikon:
Pull Request for https://github.com/joomla/joomla-cms/pull/21722

### Summary of Changes
don't send a referrer on the request

### Testing Instructions
make a request
click on the link that is send to you
check the http header of that page with the F12 developer tools.

### Expected result
On this specific page we set the Referrer-Policy and set it to no-referrer so no referrer is send.

### Actual result
We do not send that policy.

### DIFF
--- a/components/com_privacy/controller.php
+++ b/components/com_privacy/controller.php
@@ -38,6 +38,12 @@ class PrivacyController extends JControllerLegacy
 			return $this;
 		}
 
+		// Make sure we don't send a referer
+		if ($view === 'request')
+		{
+			JFactory::getApplication()->setHeader('Referrer-Policy', 'no-referrer', true);
+		}
+
 		return parent::display($cachable, $urlparams);
 	}
 }


administrator/components/com_privacy/controller.php
```diff
@@ -93,6 +93,12 @@ public function display($cachable = false, $urlparams = array())
 
 					return $this;
 				}
+
+				// Make sure we don't send a referer
```
mbabker: This one's not necessary.  You'll never have the plaintext token in the admin data, at best you have the bcrypt encoded token and we don't even have a token field (hidden or visible) in the form.

components/com_privacy/controller.php
```diff
@@ -38,6 +38,12 @@ public function display($cachable = false, $urlparams = array())
 			return $this;
 		}
 
+		// Make sure we don't send a referer
+		if ($view === 'request')
```
mbabker: This should actually be the confirm and remind views.  Again, the token is never present in the request form.

administrator/components/com_privacy/controller.php
alikon: removed

components/com_privacy/controller.php
alikon: ok, confirm and remind views

